PR #48, opened 2018-11-16
Multi-event Batches

bartelink:
This PR moves the Access Strategy from it's present approach of having a special case "Index page" (which was arrived at as a spike impl of a performance improvement for the Favorites test) to a more uniform layout which achieves at least the following objectives
- no special case documents needing to be ignored when propagating to other systems
- both the WIP and normal pages now equally hold events, with no waste
- reduce intrinsic contention when lots of writers are appending to the same stream by confining writes to a single document in the bulk of cases (as opposed to current scheme which always touches N+1 docs to write N events every time)
- reduce Request Charges and roundtrips by having fewer documents in the first instance
- allow competing writers to share the events array, allow them to naturally "catch up" by reading events not yet fed into their projections from the index page)

One key change herein is that we now have two indices (aside from `p`) - `i` is joined by `n` (which corresponds to the `i` of the next batch (which, on the `-1` / _Tip_ batch is the `expectedVersion`) - this does not seem to have affected insert costs

Note while some query costs have gone up, as noted, the bulk of the tests triggering this are artificially inducing the very extra roundtrips and bandwidth that this change will ultimately reduce.

### DIFF
--- a/src/Equinox.Cosmos/Cosmos.fs
+++ b/src/Equinox.Cosmos/Cosmos.fs
@@ -52,9 +52,6 @@ type [<NoEquality; NoComparison; JsonObject(ItemRequired=Required.Always)>]
         /// as it will do: 1. read 2. merge 3. write merged version contingent on the _etag not having changed
         [<JsonProperty(DefaultValueHandling=DefaultValueHandling.Ignore, Required=Required.Default)>]
         _etag: string
-        /// When we encounter the Tip, we're interested in the 'real i', which is kept in -_i for now
-        [<JsonProperty(DefaultValueHandling=DefaultValueHandling.Ignore, Required=Required.Default)>]
-        _i: int64
 
         /// base 'i' value for the Events held herein
         i: int64 // {index}
@@ -69,7 +66,7 @@ type [<NoEquality; NoComparison; JsonObject(ItemRequired=Required.Always)>]
     /// If we encounter the tip (id=-1) doc, we're interested in its etag so we can re-sync for 1 RU
     member x.TryToPosition() =
         if x.id <> Tip.WellKnownDocumentId then None
-        else Some { index = x._i+x.e.LongLength; etag = match x._etag with null -> None | x -> Some x }
+        else Some { index = x.i+x.e.LongLength; etag = match x._etag with null -> None | x -> Some x }
 /// A single event from the array held in a batch
 and [<NoEquality; NoComparison; JsonObject(ItemRequired=Required.Always)>]
     BatchEvent =
@@ -108,7 +105,7 @@ and [<NoEquality; NoComparison; JsonObject(ItemRequired=Required.Always)>]
         _etag: string
 
         /// base 'i' value for the Events held herein
-        _i: int64
+        i: int64
 
         /// Events
         e: BatchEvent[]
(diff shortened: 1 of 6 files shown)


src/Equinox.Cosmos/Cosmos.fs
```diff
@@ -277,53 +292,157 @@ module Log =
...
 module private Write =
     let [<Literal>] sprocName = "EquinoxIndexedWrite"
+    let [<Literal>] sprocBody = """
+    function indexedWrite(batch) {
+        var response = getContext().getResponse();
+        var collection = getContext().getCollection();
+        var collectionLink = collection.getSelfLink();
+
+        if (batch === undefined || batch==null) {
+            response.setBody({ etag: null, i: -2, conflicts: null });
```
bartelink: maybe just throw ?

src/Equinox.Cosmos/Cosmos.fs
```diff
@@ -277,53 +292,157 @@ module Log =
...
+                        // If the document is found, update it.
+                        // There is no need to check for a continuation token since we are querying for a single document.
+                        tryUpdate(documents[0], false);
+                    } else if (responseOptions.continuation) {
+                        // Else if the query came back empty, but with a continuation token; repeat the query w/ the token.
+                        // It is highly unlikely for this to happen when performing a query by id; but is included to serve as an example for larger queries.
+                        tryQueryAndUpdate(responseOptions.continuation);
+                    } else {
+                        // Else the snapshot does not exist; create snapshot
+                        var doc = {p:batch.p, id:"-1", i:-1, e:[], c:[]};
```
bartelink: `i` should not ever be -1 - better to not set it

src/Equinox.Cosmos/Cosmos.fs
```diff
@@ -277,53 +292,157 @@ module Log =
...
+                        tryUpdate(doc, true);
```
bartelink: comment on the `true` please

src/Equinox.Cosmos/Cosmos.fs
```diff
@@ -277,53 +292,157 @@ module Log =
...
+            }
+
+            // The kernel function
+            function tryUpdate(doc, isCreate) {
+
+                // DocumentDB supports optimistic concurrency control via HTTP ETag.
+                var requestOptions = {etag: doc._etag};
+
+                // Step 1: Update snapshot document's events
+                Array.prototype.push.apply(doc.e, batch.e);
```
bartelink: will this add N events to the tail, or add an array to the tail ? needs to be former

bartelink:
@jinglegit for avoidance of doubt, I'm working on the JS atm, in line with comments above

src/Equinox.Cosmos/Cosmos.fs
bartelink: The incoming batch is intended to be used directly in this case, replacing with that

src/Equinox.Cosmos/Cosmos.fs
bartelink: Verified this is def the best solution

src/Equinox.Cosmos/Cosmos.fs
bartelink: changed to throw

bartelink:
OK, I know lots more JS now; there are some NREs, presumably due to off by one errors. Also some of the tests are failing due to needing to be adjusted to fit the new representation (the test cases assume one event per item, and max batchSize events per Batch, which needs to be multiplied by the page size where one is in force etc)

PR #55, opened 2018-11-29
Facilitate running load tests inside WebHost

bartelink:
This is definitely early-days WIP atm. but plan is:
- [x] Pull out store-neutral building logic
- [x] Pull out test building logic (yielding good results, but pulling on the thread is taking longer than one might expect)
- [x] Make Web host work (it has TODOs on `cosmos` reflecting work I have not done yet)
- [x] Allow CLI to run against `Web` Host (as an other option alongside `Cosmos`, `Es` and `Mem`


bartelink:
@voronoipotato FYI there's a copy of the `samples/Store/Web` on `master` now which compiles and functions as a web api (but no front end). From here, in theory all we need is for someone to 'just' build a nice Fable-based FE (and possibly polish the Web layer a bit more - the wiring is not idiomatic aspnetcore as I've little experience with it)

In real live, my next step on the demos/docs side of the house is probably to do a https://www.todobackend.com port based on https://github.com/ChristianAlexander/dotnetcore-todo-webapi/blob/master/src/TodoWebApi/Controllers/TodosController.cs as this gives me a frontend (and I can build a CLI-driven benchmark which meets my needs in terms of being able to optimize perf).

cc @mateusz-kula

Issue #64, opened 2018-12-11
Add dotnet global tool edition of CLI


bartelink:
While the CLI needs to remain runnable from full FW in order to be able to run benchmarks in full framework (so we can't just go in and change the CLI to [only] be a dotnet tool), the provisioning and running facilities are environment independent. We should thus wra the CLI as a global tool (and make that be available via nuget)

PR #69, opened 2018-12-18
Equinox cli as .net tool package

enricosada:
fix https://github.com/jet/equinox/issues/64

- pack `Equinox.Cli` as .NET Tool
- bundle the `net461` executable in the package, in the dir `tools/net461/any`

## .net core tool

To test is locally, after `dotnet pack build.proj`

```
dotnet tool install --add-source "bin\nupkg" --tool-path bin2 --version 0.0.1 Equinox.Cli
```

and

```
bin2\eqx --help
```

use `-g` instead of `--tool-path` to install globally

## .net executable

it's inside the nupkg, in the `tools/net461/any` directory

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -9,6 +9,8 @@
     <PackageLicenseUrl>https://github.com/jet/equinox/blob/master/LICENSE</PackageLicenseUrl>
     <Copyright>Copyright © 2016-8</Copyright>
 
+    <RepoDir>$([System.IO.Path]::GetFullPath("$(MSBuildThisFileDirectory)"))</RepoDir>
+
     <IsOSX Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::OSX)))' == 'true'">true</IsOSX>
 
     <!-- SourceLink related properties https://github.com/dotnet/SourceLink#using-sourcelink -->
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -30,7 +30,7 @@ jobs:
       VersionSuffix: '$(VersionSuffix)'
   - task: PublishBuildArtifacts@1
     inputs:
-      pathtoPublish: 'bin'
+      pathtoPublish: 'bin/nupkg'
       artifactName: 'nupkgs'
   - task: NuGetCommand@2
     inputs:
@@ -39,7 +39,7 @@ jobs:
       publishFeedCredentials: 'Jet-MyGet'
       versioningScheme: byEnvVar
       versionEnvVar: Version
-      packagesToPush: 'bin/*.nupkg;bin/*.symbols.nupkg'
+      packagesToPush: 'bin/nupkg/*.nupkg'
 
 - job: Linux
   pool:
--- a/build.proj
+++ b/build.proj
@@ -6,7 +6,7 @@
     <Cfg>--configuration Release</Cfg>
 
     <ThisDirAbsolute>$([System.IO.Path]::GetFullPath("$(MSBuildThisFileDirectory)"))</ThisDirAbsolute>
-    <PackOptions>-o $(ThisDirAbsolute)bin --version-suffix "$(VersionSuffix)"</PackOptions>
+    <PackOptions>-o $(ThisDirAbsolute)bin/nupkg --version-suffix "$(VersionSuffix)"</PackOptions>
 
     <TestOptions>--logger:trx</TestOptions>
     <!-- disable known test failures on mono -->
--- a/build.proj
+++ b/build.proj
@@ -14,7 +14,8 @@
   </PropertyGroup>
 
   <Target Name="Pack">
-    <Exec Command="dotnet pack cli/Equinox.Cli $(Cfg) $(PackOptions)" />
+    <Exec Command='dotnet publish cli/Equinox.Cli $(Cfg) -f net461 -o "$(RepoDir)/bin/equinox-cli/net461" ' />
+    <Exec Command="dotnet pack cli/Equinox.Cli $(Cfg) $(PackOptions) /p:PackAsTool=true" />
     <Exec Command="dotnet pack src/Equinox $(Cfg) $(PackOptions)" />
     <Exec Command="dotnet pack src/Equinox.Codec $(Cfg) $(PackOptions)" />
     <Exec Command="dotnet pack src/Equinox.EventStore $(Cfg) $(PackOptions)" />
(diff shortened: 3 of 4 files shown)


Directory.Build.props
```diff
@@ -9,6 +9,8 @@
     <PackageLicenseUrl>https://github.com/jet/equinox/blob/master/LICENSE</PackageLicenseUrl>
     <Copyright>Copyright © 2016-8</Copyright>
 
+    <RepoDir>$([System.IO.Path]::GetFullPath("$(MSBuildThisFileDirectory)"))</RepoDir>
```
bartelink: Can you coalesce with ThisDirAbsolute ?

bartelink:
💭 It would be nice for us to do a dry-run of the release process:
a) tagging + release notes + a (-Pre) nuget (from `master`) of this as-is
b) then adding a `readme.md` overview in a PR and doing a release nuget

(It doesn't have to happen on this PR, but it's a _lot_ smaller than #42 is going to be)

Directory.Build.props
enricosada: yes, but the issue is that i need it in a fsproj, the `cli\Equinox.Cli\Equinox.Cli.fsproj` like

```xml
    <Content Include="$(RepoDir)/bin/equinox-cli/net461/*">
```

if i wrote it as `$(ThisDirAbsolute)/bin/equinox-cli/net461/*` and it's confusing, because is not that file dir, but the repo dir

enricosada:
@bartelink to dryrun, feel free to change the release pipeline feed, @michaelliao5 know how to do that, to push to another feed (there is a myget test feed for that).

enricosada:
> Should/can we alias the Equinox.Cli so one can do a `dotnet equinox <command line>`, or am I missing something?

so, if you want `dotnet equinox` you need to change the name of executable to `dotnet-equinox` (it's just a convention)

but ihmo, is better to just call it `equinox`, so can be invoked as `equinox --help`, not `dotnet equinox --help`
the `dotnet something` is useful if you want to mimic also the other options (`-f` , `-c` etc), not for normal execuables

bartelink:
Ah, I was missing lots of things! I thought a tool could not be run without `dotnet` prefixing.

The problem is there's a lib called `Equinox`, and that name makes sense.

I'm now torn between renaming `cli/Equinox.Cli` to `src/dotnet-equinox` or just leaving it as-is now, btu am also open to other ideas 😕. 

>feel free to change the release pipeline feed

I'm interested in a dry-run of it being on nuget.org (I'll do a quick sanity check from myget first)

Directory.Build.props
bartelink: Maybe remove L28-30 and use RepoDir below that ? I could argue for the duplication staying too so will leave it up to your taste - just wanted to make sure the duplication was intentional and not an oversight

enricosada:
> I'm interested in a dry-run of it being on nuget.org (I'll do a quick sanity check from myget first)

it will be the same package as deployed in myget now, just on nuget.org
i already tried the release pipeline with a dev feed, and @michaelliao5 changed the destination to nuget.org now.
So no need to change the release pipeline, just check the build nupkg in myget (or inside the azure pipeline artifact https://dev.azure.com/jet-opensource/opensource/_build/results?buildId=606&view=results )


> I'm now torn between renaming cli/Equinox.Cli to src/dotnet-equinox or just leaving it as-is now, btu am also open to other ideas 😕.

@ruben you dont need to rename the project, is fine as it is, so the package name is correct.
just set the `AssemblyName` msbuild property, like

```xml
   <AssemblyName>equinox</AssemblyName>
```

ihmo:

- as cli name `Equinox.Cli` is orrible, and hard to digit and redundant.
- `dotnet equinox` doesnt make sense, because it doesnt mimic a dotnet command, is just an console app
- `equinox` can be ok, i have done a search and there isnt one in ubuntu packages either

bartelink:
OK, that makes sense. It leaves two outstanding minor puzzlers though
- the CLI's package name will be `Equinox.Cli`
- it's in a `cli` subdirectory

1. I'm not sure if changing just the assembly name and having that be out of sync with everything else will help the average newcomer
2. I hate the `.Cli` too - It was called Equinox.Bench before, but it's clear it will be multi-purpose for the foreseeable
3. As stated before, I don't want to change the Equinox package's name

🤔 I assume there is no prior art for `Cli` in a name, and similar for a `cli` directory
💭 should/can the folder become `tools` and the project `Equinox.Tool`, but with the exe always called Equinox.EXE (even if it refences Equinox.DLL - for all imaginable hosting scenarios?)

enricosada:
package name (`PackageId`, by defualt the project name) and the output executable name (`AssemblyName`, by default the project name) shouldnt be the same.

feel free to choose dir, packageid and exename.

if you have a library called `Equinox.dll` maybe is an issue call the exename `equinox`, because collide on win (the exe assemblyu name will be `equinox.dll`).
maybe use a shortname? `eqx` or something like that?

bartelink:
`eqx.exe` is probably a good solution for the name of the EXE

that leaves the `cli` folder and `.Cli` part of the nuget package and project - I'm thinking a `tools` folder and calling it `Equinox.Tool` in terms of folder/fsproj/nuget id would work ? Unless there's more than me liking that though, there's probably no point in going to the trouble of a rename

enricosada:
directory name and package name doesnt matter, it's just how ppl install it.

```bash
dotnet tool install -g Equinox.Cli
# vs
dotnet tool install -g Equinox.Tool
```

it's the same, the .net tools nupkg have a flag in the metadata (`PackageType=Tool`) so can be filtered in the gallery, is not done using the name of the package.

I'll rename to `eqx` so, and leave in `.Cli` dir for now, we can rename in another PR

bartelink:
Thanks for taking the time to add the detailed instructions - I took if for a spin and discovered two things:

1) need to add a PackageId or the nupkg becomes eqx.nupkg (surprising, I'd argue it should be an SDK bug?) 
2) we should either add a `PrivateAssets="all"` to TodoBackend or remove it from the others (I'm guessing the latter would work?)

Happy for it to be merged after that, it looks great, thanks!

bartelink:
Also there's a stray `PrivateAssets=All` in Infrastructure.froj (grep for it).

If you remove that, the following can be removed from the Cli project

    <ProjectReference Include="..\..\src\Equinox.MemoryStore\Equinox.MemoryStore.fsproj" />

bartelink:
https://github.com/jet/equinox/tree/enricosada-eqx_clitool

bartelink:
About to squash-merge - will apply my tweaks as an amend shortly afterwards

PR #70, opened 2018-12-19
Use EventStore.Client 5.0.0 to remove EventStore.Client.NetCore dependency on netstandard2.0

bartelink:
To run:

    # requires admin privilege
    cinst eventstore-oss -y # where cinst is an invocation of the Chocolatey Package Installer on Windows
    # run as a single-node cluster to allow connection logic to use cluster mode as for a commercial cluster
    & $env:ProgramData\chocolatey\bin\EventStore.ClusterNode.exe --gossip-on-single-node --discover-via-dns 0 --ext-http-port=30778

    # NETCORE
    & dotnet run -c Release -f netcoreapp2.1 -p tools/Equinox.Tool -- run -f 2000 -d 5 -C es

    # FULL FW
    dotnet pack -c Release .\build.proj
    & .\tools\Equinox.Tool\bin\Release\net461\eqx.exe run -f 2500 -C -d 5 es


bartelink:
While it's not a breaking change, it should be noted per @jageall that both https://github.com/EventStore/ClientAPI.NetCore and the `netstandard2.0` variant in the `5.0.0-rc3` do not internalize dependencies (notably protobuf, but also json.net), which has been observed to cause issues in some cases

PR #72, opened 2018-12-20
Rename cli/Equinox.Cli to tools/Equinox.Tool

bartelink:
This naming fits a teeny bit better alongside `Equinox.Templates`, and allows us to add other tools ... like Projectors ;)


bartelink:
I had tried to omit the net461 build from the `Equinox.Tool.nupkg`, but unfortunately it was still too big.

PR #75, opened 2018-12-21
Use MinVer

bartelink:
This PR, a rerun of #71, integrates [MinVer](https://github.com/adamralph/minver) 

HT: this is based on https://github.com/justeat/JustSaying/pull/465, which @adamralph kindly sent me a link to

To test locally (the Azure Pipelines effectively invokes it this way): `dotnet pack /p:build_pr=46 /p:build_id=1000 build.proj`

TODO
- [ ] simplify the wiring in the `azure-pipelines.yml` to pass `BUILD_ID` and `BUILD_PR`
- [ ] not seeing VERSIONINFO on the dlls

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,6 +1,5 @@
 <Project ToolsVersion="15.0">
   <PropertyGroup>
-    <VersionPrefix>0.0.1</VersionPrefix>
     <Authors>@jet @bartelink @eiriktsarpalis and contributors</Authors>
     <Company>Jet.com</Company>
     <Description>Composable high performance event sourcing componentry</Description>
@@ -21,7 +20,8 @@
     <EnableSourceLink Condition=" '$(OS)' != 'Windows_NT' AND '$(MSBuildRuntimeType)' != 'Core' ">false</EnableSourceLink>
 
     <!-- suppress false positive warning FS2003 about invalid version of AssemblyInformationalVersionAttribute -->
-    <NoWarn>FS2003</NoWarn>
+    <!-- Supress NU5105 triggerd by MinVer height e.g.: pr43-rc1.2: The package version '<X>' uses SemVer 2.0.0 or components of SemVer 1.0.0 that are not supported on legacy clients. Change the package version to a SemVer 1.0.0 string. If the version contains a release label it must start with a letter. This message can be ignored if the package is not intended for older clients. -->
+    <NoWarn>$(NoWarn);FS2003;NU5105</NoWarn>
   </PropertyGroup>
 
   <!-- Workaround for https://github.com/xunit/xunit/issues/1357 -->
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -3,4 +3,18 @@
   <Target Name="VSTestIfTestProject">
     <CallTarget Targets="VSTest" Condition="'$(IsTestProject)' == 'true'" />
   </Target>
+  <Target Name="ComputePackageVersion" AfterTargets="MinVer" Condition="'$(BUILD_PR)' != ''" >
+    <PropertyGroup>
+      <PackageVersion>$(MinVerMajor).$(MinVerMinor).$(MinVerPatch)-pr.$(BUILD_PR)</PackageVersion>
+      <PackageVersion Condition="'$(MinVerPreRelease)' != ''">$(PackageVersion).$(MinVerPreRelease)</PackageVersion>
+      <PackageVersion Condition="'$(MinVerBuildMetadata)' != ''">$(PackageVersion)+$(MinVerBuildMetadata)</PackageVersion>
+      <Version>$(PackageVersion)</Version>
+    </PropertyGroup>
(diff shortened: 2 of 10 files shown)


Directory.Build.props
```diff
@@ -21,7 +20,8 @@
     <EnableSourceLink Condition=" '$(OS)' != 'Windows_NT' AND '$(MSBuildRuntimeType)' != 'Core' ">false</EnableSourceLink>
 
     <!-- suppress false positive warning FS2003 about invalid version of AssemblyInformationalVersionAttribute -->
-    <NoWarn>FS2003</NoWarn>
+    <!-- Supress NU5105 triggerd by MinVer height e.g.: pr43-rc1.2: The package version '<X>' uses SemVer 2.0.0 or components of SemVer 1.0.0 that are not supported on legacy clients. Change the package version to a SemVer 1.0.0 string. If the version contains a release label it must start with a letter. This message can be ignored if the package is not intended for older clients. -->
+    <NoWarn>FS2003;NU5105</NoWarn>
```
adamralph: It's better to do `<NoWarn>$(NoWarn);FS2003;NU5105</NoWarn>` otherwise you risk overwriting values which have been set elsewhere.

src/Equinox.Codec/Equinox.Codec.fsproj
```diff
@@ -15,6 +15,8 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0-beta-63127-02" PrivateAssets="All" />
+    <PackageReference Include="MinVer" Version="1.0.0-beta.2+build.172" />
```
adamralph:
```suggestion
    <PackageReference Include="MinVer" Version="1.0.0-beta.2" />
```

Build metadata is not part of the version: https://www.nuget.org/packages/MinVer/1.0.0-beta.2

(Same comment applies to all the other package references.)

azure-pipelines.yml
```diff
@@ -5,14 +5,15 @@ jobs:
...
-      $isPR = "$env:SYSTEM_PULLREQUEST_PULLREQUESTNUMBER" -ne ""
-      $versionSuffix = if ($isTag) { "" } else { if ($isPR) { $versionSuffixPR } else { $versionSuffixBRANCH } };
-      Write-Host "##vso[task.setvariable variable=VersionSuffix]$versionSuffix";
+      $buildId = $env:BUILD_BUILDNUMBER.PadLeft(7, '0')
+      if ("$env:SYSTEM_PULLREQUEST_PULLREQUESTNUMBER" -ne "") {
+        $meta = "$buildId-pr$($env:SYSTEM_PULLREQUEST_PULLREQUESTNUMBER)"
+      } else {
+        $branchName = "$env:BUILD_SOURCEBRANCHNAME".Replace("_","");
+        $meta = "$buildId-$branchName";
+      }
```
adamralph: Maybe I'm missing something but it seems like `$meta` is only being assigned, and its value is not actually used anywhere.

Directory.Build.props
```diff
@@ -21,7 +20,8 @@
     <EnableSourceLink Condition=" '$(OS)' != 'Windows_NT' AND '$(MSBuildRuntimeType)' != 'Core' ">false</EnableSourceLink>
 
     <!-- suppress false positive warning FS2003 about invalid version of AssemblyInformationalVersionAttribute -->
-    <NoWarn>FS2003</NoWarn>
+    <!-- Supress NU5105 triggerd by MinVer height e.g.: pr43-rc1.2: The package version '<X>' uses SemVer 2.0.0 or components of SemVer 1.0.0 that are not supported on legacy clients. Change the package version to a SemVer 1.0.0 string. If the version contains a release label it must start with a letter. This message can be ignored if the package is not intended for older clients. -->
```
adamralph: FTR it's not just the height that triggers this. It's also the build metadata, which was introduced in SemVer 2.0.

bartelink: yeah - we had not leant on that (we should now that I remember)

Directory.Build.targets
```diff
@@ -3,4 +3,18 @@
   <Target Name="VSTestIfTestProject">
     <CallTarget Targets="VSTest" Condition="'$(IsTestProject)' == 'true'" />
   </Target>
+  <Target Name="ComputePackageVersion" AfterTargets="MinVer" Condition="'$(BUILD_PR)' != ''" >
+    <PropertyGroup>
+      <PackageVersion>$(MinVerMajor).$(MinVerMinor).$(MinVerPatch)-pr.$(BUILD_PR)</PackageVersion>
```
adamralph: FTR this will also trigger NU1505 because of the dot after `pr`. Not that it matters (I don't belive it's worth supporting those older clients) but it's another instance where it's not just height that triggers it. It may be worth altering the comment before the suppression of NU1505.

bartelink: Good point, will do (elsewhere)

PR #77, opened 2018-12-24
Make minver a private asset

eiriktsarpalis:
Minver is currently leaking as a dependency in the eqx libraries, causing compile time errors when referenced from libraries that are using the old project system.


bartelink:
Thanks (dammit - I actually went to inspect whether this was transitive and decided [incorrectly] it wasn't - reason is that NuGet package explorer chokes on the source integration)

PR #91, opened 2019-02-05
Update Projection to use `Confluent.Kafka.1.0.0-beta2`

michaelliao5:
This PR updates the `Confluent.Kafka` reference from 0.11.4 to the above.

The core of the changes here are based on [the examples](https://github.com/confluentinc/confluent-kafka-dotnet/tree/master/examples)

Reduction in number of config settings is from a mix of spotting redundancy based on [the docs](https://github.com/edenhill/librdkafka/blob/master/CONFIGURATION.md) and in reading the xmldoc for the ConsumerConfig/ClientConfig/ProducerConfig classes in the `Confluent.Kafka` library

_It should be noted that moving to 0.11.5 or 0.11.6 also necessitate similar changes to the shutdown logic (which sometimes can be reduced by changing the number of consumers to 1 in the happy path test)_.

### DIFF
--- a/src/Equinox.Projection.Kafka/Kafka.fs
+++ b/src/Equinox.Projection.Kafka/Kafka.fs
@@ -25,15 +25,15 @@ module Config =
 
 // NB we deliberately wrap all types exposed by Confluent.Kafka to avoid needing to reference librdkafka everywhere
 [<NoComparison>]
-type KafkaDeliveredMessage internal (message : DeliveryReportResult<string,string>) =
+type DeliveryReport internal (message : DeliveryReportResult<string,string>) =
     member internal __.UnderlyingMessage = message
     static member TryWrap (msg : DeliveryReportResult<string,string>) =
         if msg.Error.IsError then
             let errorMsg = sprintf "Error on message topic=%s code=%O reason=%s" msg.Topic msg.Error.Code msg.Error.Reason
             Choice2Of2 (Exception errorMsg)
         else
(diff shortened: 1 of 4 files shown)


src/Equinox.Projection.Kafka/Kafka.fs
```diff
@@ -294,49 +228,56 @@ type KafkaProducer private (log: ILogger, producer : Producer<string, string>, t
...
+                for fmt in fmtTopicPartitionOffsetErrors cos.Offsets do log.Information("consumer_committed_offsets|{offsets}{oe}", fmt, fmtError cos.Error))
             { new IDisposable with member __.Dispose() = for d in [d1;d2;d3;d4;d5;d6] do d.Dispose() }
 
-    let mkConsumer (kvps, topics : string seq) keyDeserializer valueDeserializer =
-        let consumer = new Consumer<'Key, 'Value>(kvps, keyDeserializer, valueDeserializer)
-        let _ = consumer.OnPartitionsAssigned.Subscribe(fun m -> consumer.Assign m)
-        let _ = consumer.OnPartitionsRevoked.Subscribe(fun _ -> consumer.Unassign())
+    let mkConsumer (kvps : seq<KeyValuePair<string,string>>, topics : string seq)  =
+        let consumer = new Consumer<'Key, 'Value>(kvps)
+        //let _ = consumer.OnPartitionsAssigned.Subscribe(fun m -> consumer.Assign m)
```
bartelink: can these be reinstated?

Issue #104, opened 2019-03-01
Confront consumers with opt-out from caching in Equinox.Cosmos StreamResolver


bartelink:
As flagged by @jakzale, its critical that consumers be confronted early in the experience with decision making processes around caching and/or unfolds and/or out-of-collection snapshots early in the game.

🤔 we could simply make the caching argument not be optional, at least until #61 is fleshed out more

This also could benefit from more docs/articles and/or tutorials and/or samples ref #57

PR #107, opened 2019-03-04
Remove optionality from Cosmos caching

bartelink:
Adds friction per #104 to the `CosmosResolver` API to ensure people have to go out of their way to avoid caching


bartelink:
@jakzale Any wording improvements or other suggestions appreciated

Issue #109, opened 2019-03-08
Cosmos: Harden and merge `tip-isa-Batch` branch


bartelink:
There's a `tip-isa-batch` branch, which generalizes the storage scheme to provide even greater efficiency in terms of read and write costs.

This work was shelved for prioritisation purposes in order to avoid having to make projectors deal cleanly with seeing an event multiple times.

Given a projector that does some practical de-duplication of events, its pretty feasible to re-introduce this storage optimization; the performance and RU cost reductions can be inspected by using `Equinox.Tool run` and/or `dotnet new eqxtestbed` against said branch.

As noted in #108, this storage generalization also provides benefits wrt competing writer scenarios.

bartelink:
Removed milestone to reflect that, while this is unlikely to ever make it to a 2.x release (targeting Azure.Cosmos V2), it's also not guaranteed to make a 3.x release as the Archival/Trimming facilities in #226 will likely yield a better overall solution. The other reason it's not a no-brainer to add back in is that the write amplification effects of having to read, write, and then have the CFPs read the redundant events are such that it requires a very specific combination of things (e.g. small aggregate bytes of events per aggregate) in order for this to be the best solution

bartelink:
More reflections: Learnings and realizations from implementing Archival+Pruning actually make the case _stronger_ for implementing this:
- The archiver has the opportunity and incentive to spend more time packing (events in tip and/or compressing event bodies) when writing to a secondary - the write amplification considerations don't apply to the same extent as they do on primary

PR #110, opened 2019-03-12
Equinox.Cosmos: Reinstate Tip isa Batch

bartelink:
The CosmosDb Stored Procedure in  1.x and 2.x guarantees to to never to mutate a document.

This was considered to be a critical guarantee on which internal projection systems would depend.

I'm of the opinion that the At-Least-Once-delivery characteristics of anything consuming from the CosmosDb Change Feed imply any projector can yield repeat sequences that are no different to those one might anticipate were mutation of events taking place.

As a result, the main thing that's necessary to consider exposing this facility is to ensure the following are adhered to:
- [ ] Downstreams should not exclude documents with `id="-1"`
- [ ] Downstream projectors should consider removing duplicate events where they are absolutely certain they have already projected an item in the current sessions (as noted above, consumers should already be able to handle partial resends of earlier events, _including patterns such as 012314345678129, but this change will certainly make repeats more prevalent_(so implementing this may be avoidable))


bartelink:
Resolves #109

bartelink:
Superseded by #251

Issue #120, opened 2019-04-03
`eqx init` should support provisioning at database level


bartelink:
There are various scenarios where it makes sense to allocate RUs across collections when managing notifications via the ChangeFeedProcessor

Relevant Docs:
- https://docs.microsoft.com/en-us/azure/cosmos-db/set-throughput
- https://azure.microsoft.com/en-us/blog/sharing-provisioned-throughput-across-multiple-containers-in-azure-cosmosdb/

PR #121, opened 2019-04-03
Revamp `eqx init` to address DB level allocations and adjustments resolves #112 #120

### LINKED ISSUE #112: `eqx init -ru <value>` should change allocation for existing collections
- needs to log `WRN` if not being applied immediately

bartelink:
- [x] #112 re being able to re-run commands to adjust/validate provisioning idempotently
- [x] #120 re being able to provision at database level
- [x] Tidy
- [ ] Document reasoning


bartelink:
Someone will bug me for docs in due course; ideally we'll have more learnings regarding the applicability of the technique at the time such typing gets triggered ;)

PR #125, opened 2019-05-05
Split out Equinox.Projection.Codec

bartelink:
Work on the `eqxsync` template in https://github.com/jet/dotnet-templates/pull/16 has yielded a replacement for `FeedValidator`; this PR reorgs in advance of that.

- [x] formalize `Equinox.Projection.StreamItem` to represent an event from a stream that's being projected
- [x] splits an `Equinox.Projection.Codec` NuGet out from `Equinox.Projection` (this was the name used for `Equinox.Projection` until `2.0.0-preview3`;  it's wot clear that we have enough concepts to warrant doing the split)


bartelink:
(cherry-picked merge to `master`, which was the original plan)

PR #134, opened 2019-05-21
Remove TTLs, indexing from Aux config

bartelink:
This PR cleans up the `initaux` indexing behavior with the aim of maximizing efficiency and correctness :-
- removing `DefaultTimeToLive` on order to avoid risk of Lease and/or other documents used by either the present or future versions of the CFP lib being removed (used by an internal alternate projector system; use of that system will now now necessitate applying such configuration manually)
- removing indexing (now turned to `Automatic=false, IndexingMode = None`)


bartelink:
See related questions in https://github.com/Azure/azure-documentdb-changefeedprocessor-dotnet/issues/142

PR #135, opened 2019-05-29
Fix EventStore Connection Logic wrt Slave/Random nodes on cluster connections

bartelink:
This code, derived from a fix by @asetda (Aset Dauletbaev) to a downstream version fixes the fact that the gossip-based Discovery profiles don't pick up the `.PreferRandomNode()`/`.PreferSlaveNode()` preference on EventStore's `ConnectionSettings`

There's also a fix for a typo which means that the `PreferSlave` semantics would never have worked even in the `Uri` case (this was not present in the downstream repo)...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ The `Unreleased` section name is replaced by the expected version of next releas
 ### Removed
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


bartelink:
Cause was identified and fixed by @asetda

PR #143, opened 2019-07-02
Change env vars from EQUINOX_KAFKA* -> PROPULSION_*

bartelink:
Same change is happening in https://github.com/jet/dotnet-templates to reflect that projection is not tied to Equinox, and ease confusion badging everything EQUINOX can bring.

(NB this information will likely eventually move to the propulsion API as part of https://github.com/jet/propulsion/issues/6)


bartelink:
Should be rebased/merged after #138

PR #144, opened 2019-07-02
Port to Azure.Cosmos / v3 SDK

bartelink:
- [x] Need to figure out best way to invoke stored proc with 3 args, of which one is a large object https://github.com/Azure/azure-cosmos-dotnet-v3/issues/509
- [x] Need to figure out why 9f94b6d88a65f691e0a6e38584dbfb731b44449c is required (some queries got more expensive) https://github.com/Azure/azure-cosmos-dotnet-v3/issues/722
- [ ] port `eqx dump` TODO (deferring for now - V4 port will include it)


bartelink:
It passes; there's a TODO or two, but in principle this can get merged into `master` after we declare master to be `3.x-preview`-era
(it may make sense to introduce an `Equinox.Cosmos2` nuget with identical code using `Microsoft.Azure.Documents` for backcompat)

bartelink:
Merging to checkpoint this work.

There's in-flight work to migrate `Equinox.Cosmos` to use the `V4` `Azure.Cosmos` SDK - the likelihood is that there won't thus be a long-lived version of `Equinox.Cosmos` targettng the V3 client, i.e. the first 3.0.0-beta releases will likely use the V4 SDK

PR #152, opened 2019-08-23
Expose aggregate Version on Equinox.Stream.QueryEx

bartelink:
When producing summaries, one often performs a `Query` but need to know the Version in question. While this can be inferred from the underlying events, factors like:
- `Equinox.Cosmos.RollingState` (which has snapshots that get revised, which means event level `Index` value is not meaningful)
- each store has its own way of representing versions etc
complicate matters.

This PR exposes a `Version` alongside the state solely for the `QueryEx` method in order to meet a need of the Summary Projection template (having the change feed triggering the app emitting a summary - knowing the version one read allows one to skip projections based on an older state in the projector)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ The `Unreleased` section name is replaced by the expected version of next releas
 ## [Unreleased]
 
 ### Added
+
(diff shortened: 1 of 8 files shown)


bartelink:
@ameier38 IIRC we had a discussion about something of this kind at one point. Would this allow you to simplify anything anywhere and/or would a minor tweak help that?

Issue #145, opened 2019-07-12
Customizing cache implementation


DSilence:
I've been researching the library and it looks really neat!

I've been wondering if there are plans to allow the customization if cache implementation for GES and CosmosDB providers.
At the moment it uses System.Runtime.Caching.MemoryCache instance. 

Microsoft.Extensions.Caching provides multiple abstractions (IMemoryCache and IDistributedCache), and the possibility to customize the implementation seems like a valuable option. E.g. in our system MemoryCache from Microsoft.Extensions.Caching is used heavily, and having multiple caching providers doesn't make much sense.

bartelink:
Thanks!

Have not felt the need to generalize in this direction but can appreciate your desire to have only one set of cache stuff to reason about.

One thing to bear in mind is that a distributed cache is questionable; the S.C.MC holds object refs and does not require serialization - a distributed cache in front of GES is pretty pointless as GES already has good caching built-in and a distributed cache in front of cosmos likely will cost as much in management overhead as it will save in RU cost.

Assuming it doesnt cause too much mess and achieves something worthwhile, would take a PR to make the wiring in of the cache provider pluggable via an optional arg on the resolver etc. (would prefer not to have two concrete impls in the actual file `EventStore.fs` and `Cosmos.fs` files though - if anything it'd be nice to define an interface in a shared place and then have your impl and the stock one). At present there are small variances in the key strategy between Cosmos and ES; the code has enough tests to make doing a spike feasible, but I wont suggest it's going to be trivial to do.

Can you explain a little about how your `MemoryCache` operates - does it hold a ref to an object and self-prune like SRCMC?

DSilence:
I do agree that the benefits of distributed caching are very questionable (and you likely have investigated into it way more than I have). Although, I do believe that  building a common abstraction may be handy as more providers are added.

As for the `MemoryCache`, overall it works in a similar manner to SRCMC:
- The self-prune mechanism based on the "cost" of the object - which has to be specified manually, as it is not the "actual" size of the object, but some abstract value specified during insert process (https://docs.microsoft.com/en-us/aspnet/core/performance/caching/memory?view=aspnetcore-2.2#use-setsize-size-and-sizelimit-to-limit-cache-size). By default, if nothing is specified, it effectively functions as an object count limit
- It doesn't support windows performance counters
- It does hold a ref to the original object
- Overall the API is somewhat nicer and there is no need to use `ConfigurationManager` to adjust it's behavior

bartelink:
[...]

Issue #157, opened 2019-09-04
Add ResolveOption.AllowStale to Resolver


bartelink:
`Resolve`s originally used the `Target` alone to signal load hints. One of these being `Target.AggregateIdEmpty`, which signals that the system should not attempt to load the stream when necessitated by a `Query` or `Transact`, but (e.g. if the client has just minted a fresh stream id) should instead work on the assumption that no events will be loaded.

This concept generalizes to having a read mode option, which allows one to specify the following:
- [x] `AssumeEmpty` - as in OP ~, check cache, but if no value held,~ assume empty stream
- [x] `AllowStale` - hint that if a cached value is present that a a read should optimistically use that state and position for a first attempt (for an execution cycle, that means leaning on the conflict detection to trigger an automatic resync iff we're on a node whose cache happens to be stale wrt the given stream - for a read, it means you're not going to be able to guarantee that you're yielding the freshest state available in the cluster)
- 🤔 if `Equinox.Cosmos` ever needs an equivalent of `Equinox.EventStore.CachingStrategy.SlidingWindowPrefixed`, a prefix could potentially be supplied via this mechanism

Issue #62, opened 2018-12-08
Support SqlStreamStore


bartelink:
[SqlStreamStore](https://github.com/SQLStreamStore/SQLStreamStore) is a SQL-focused library addressing many of the concerns Equinox does, with support for SQL Server and Postgres. Given the common EventStore (and DDD-CQRS-ES slack!)-influenced designs, there's a good chance there'd be minimal work required on either side to add an `Equinox.SqlStreamStore` adapter

Doing this work would:
- improve the breed wrt generalizing the store interfaces in Equinox (SqlStreamStore has a wealth of experience behind it)
- provide an avenue for folks looking for a hosted solutions without having to concern themselves with request charges (likely giving up some perf, but the CLI benchmark facility will tell all)

Initial implementation should probably take a dependency on a Sql LocalDb, but there's an obvious need for it to work well with Azure SQL (the CLI would likely manage the provisioning etc. see #59)

bartelink:
Hey Nat, good question. While I'm aware of Marten from the blog and a podcast or two, I'm unfamiliar with the details of what Marten does wrt event sourcing and/or whether it would mesh well. Wild guess says it brings a programming model and an opinion to the table to a greater degree than SSS does.

Clearly Postgres is a very suitable target though. (Also is it Marten Postgres-only? SQL Server / Azure SQL would be an important DB to support)

I guess the key to answering your question would be to ask whether Marten offers an easier path to integration as a first class Equinox adapter; I'd be delighted to hear anyone's insight wrt that. (other considerations are having low dependencies, and the potential to integrate detailed metrics and logging)

bartelink:
Quick scan says:
+: excellent docs, it seems the ES APIs would facilitate an easy initial implementation
-: there's a heck of a lot to Marten's codebase as it's doing all those things - that leads to lots of dependencies and code to traverse which would not be relevant for Equinox 
-: not low on dependencies ?
0: the projections stuff in the `cosmos` branch and how it will integrate into https://github.com/jet/equinox/wiki/Cosmos-Storage-Model will likely be applicable to Postgres and SQL Server too so that aspect is not beneficial)

Marten is clearly a very capable system with good docs, community and tests but my sense/bias that it's already trying to do a lot of things (and that this fact would make an integration problematic) isn't reduced by my scan

NatElkins:
[...]

Issue #61, opened 2018-12-08
Cosmos: Provide ability to separate storage of snapshots


bartelink:
The `u`nfolds mechanism, as documented in #50 provides the ability to efficiently stash relevant data facilitating reading of state based on a point read.

However, this is can be suboptimal in the cases of larger snapshot states:-
- the data gets compressed and travels each time
- writing/updating occasions read and write costs within the sync stored procedure each time
- competing writers pay all but the update case in the case of a re-run of the same transaction
- readers not interested in/unable to process a specific snapshot schema still have to pay the price to transport it
- each update goes to the changefeed triggering more read costs
- the size and cost of the snapshots is harder to separate out
- snapshots contribute to partition size limits

Thus it makes sense to do some, or all of the following in the case where there is a projector in play and there is an `aux` collection (in the case of multiple writers one might even consider making an aux collection even if you don't have a projector):
- maintain the snapshots in the `aux` collection, where one is not paying for changefeed re-reads of the same data
- reduce the write frequency for snapshots (when the write happens should be deterministic for competing writers)
- implement ability for stored proc to maintain events since snapshot in tip when not updating snapshot
- write to a nominated collection (typically aux), with a deterministic guid based on the case / event type tag of the union as a point write
- refer to the snapshot in the unfolds; automate loading of the referenced snapshot if picked by `isOrigin`
- expose the ability to customize the snapshot loading process (but batteries should be included)
  - (perhaps do a sample showing parallel reads of the Snapshots and the Tip?)

bartelink:
The designs for #145 and #157 should consider the potential overlap with this requirement.

PR #161, opened 2019-09-21
#145 Define ICache, centralize implementation in Equinox.Core

DSilence:
I've finally had some free time to hack something together. Here is the summary of changes:
1) Added to interfaces `ICache` interface (note that it's `async` based). Moved the `CacheEntry` type nearby, changed the Token.supersedes logic to be customizable. Added new `CacheEntryOptions` which is a more abstract way to configure the behavior of cache items (relative/absolute expiration).
2) Removed store-specific cache logic. 
3) Extracted `System.Runtime.Caching` integration into a separate project. The implementation remains the same.
4) Added `Microsoft.Extensions.Caching` implementation. Couldn't find a good way to customize the way it calculates object size (which wasn't terribly slow) so added a way to customize it via constructor parameter.
The distributed cache implementation is also pretty straightforward given that there is an abstraction here now. 
Looking forward to feedback!

### DIFF
--- a/samples/Infrastructure/Storage.fs
+++ b/samples/Infrastructure/Storage.fs
@@ -76,7 +76,7 @@ module Cosmos =
         let conn = connector.Connect("equinox-tool", discovery) |> Async.RunSynchronously
         let cacheStrategy =
             if cache then
-                let c = Caching.Cache("equinox-tool", sizeMb = 50)
+                let c = Equinox.Cache("equinox-tool", sizeMb = 50)
                 CachingStrategy.SlidingWindow (c, TimeSpan.FromMinutes 20.)
             else CachingStrategy.NoCaching
         StorageConfig.Cosmos (createGateway conn batchSize, cacheStrategy, unfolds, dName, cName)
@@ -135,7 +135,7 @@ module EventStore =
         let conn = connect storeLog (a.Host, heartbeatTimeout, concurrentOperationsLimit) a.Credentials operationThrottling |> Async.RunSynchronously
         let cacheStrategy =
             if cache then
-                let c = Caching.Cache("equinox-tool", sizeMb = 50)
+                let c = Equinox.Cache("equinox-tool", sizeMb = 50)
                 CachingStrategy.SlidingWindow (c, TimeSpan.FromMinutes 20.) |> Some
             else None
         StorageConfig.Es ((createGateway conn batchSize), cacheStrategy, unfolds)
--- a/src/Equinox.Core/Cache.fs
+++ b/src/Equinox.Core/Cache.fs
@@ -23,3 +23,36 @@ type CacheEntry<'state>(initialToken: StreamToken, initialState: 'state, superse
 type ICache =
     abstract member UpdateIfNewer: cacheItemOptions:CacheItemOptions -> key: string -> CacheEntry<'state> -> Async<unit>
     abstract member TryGet: key: string -> Async<(StreamToken * 'state) option>
+
+namespace Equinox
+open System.Runtime.Caching
+open Equinox.Core
+type Cache(name, sizeMb : int) =
+        let cache =
+                let config = System.Collections.Specialized.NameValueCollection(1)
+                config.Add("cacheMemoryLimitMegabytes", string sizeMb);
+                new MemoryCache(name, config)
+
+        let getPolicy (cacheItemOption: CacheItemOptions)=
+            match cacheItemOption with
+            | AbsoluteExpiration absolute -> new CacheItemPolicy(AbsoluteExpiration = absolute)
+            | RelativeExpiration relative -> new CacheItemPolicy(SlidingExpiration = relative)
+
+        interface ICache with
+
+            member this.UpdateIfNewer cacheItemOptions key entry =
+                let policy = getPolicy cacheItemOptions
+                match cache.AddOrGetExisting(key, box entry, policy) with
+                | null ->
+                    async.Return ()
+                | :? CacheEntry<'state> as existingEntry -> existingEntry.UpdateIfNewer entry
+                                                            async.Return ()
+                | x -> failwithf "UpdateIfNewer Incompatible cache entry %A" x
+
+            member this.TryGet key =
+                async.Return (
+                    match cache.Get key with
+                    | null -> None
+                    | :? CacheEntry<'state> as existingEntry -> Some existingEntry.Value
+                    | x -> failwithf "TryGet Incompatible cache entry %A" x
+                )
--- a/src/Equinox.Core/Cache.fs
+++ b/src/Equinox.Core/Cache.fs
@@ -25,34 +25,32 @@ type ICache =
     abstract member TryGet: key: string -> Async<(StreamToken * 'state) option>
 
 namespace Equinox
+
 open System.Runtime.Caching
 open Equinox.Core
+
 type Cache(name, sizeMb : int) =
         let cache =
                 let config = System.Collections.Specialized.NameValueCollection(1)
                 config.Add("cacheMemoryLimitMegabytes", string sizeMb);
                 new MemoryCache(name, config)
 
-        let getPolicy (cacheItemOption: CacheItemOptions)=
+        let getPolicy (cacheItemOption: CacheItemOptions) =
             match cacheItemOption with
             | AbsoluteExpiration absolute -> new CacheItemPolicy(AbsoluteExpiration = absolute)
             | RelativeExpiration relative -> new CacheItemPolicy(SlidingExpiration = relative)
 
         interface ICache with
-
-            member this.UpdateIfNewer cacheItemOptions key entry =
+            member this.UpdateIfNewer cacheItemOptions key entry = async {
                 let policy = getPolicy cacheItemOptions
                 match cache.AddOrGetExisting(key, box entry, policy) with
-                | null ->
-                    async.Return ()
+                | null -> ()
                 | :? CacheEntry<'state> as existingEntry -> existingEntry.UpdateIfNewer entry
-                                                            async.Return ()
-                | x -> failwithf "UpdateIfNewer Incompatible cache entry %A" x
+                | x -> failwithf "UpdateIfNewer Incompatible cache entry %A" x }
 
-            member this.TryGet key =
-                async.Return (
+            member this.TryGet key = async {
+                return
                     match cache.Get key with
                     | null -> None
                     | :? CacheEntry<'state> as existingEntry -> Some existingEntry.Value
-                    | x -> failwithf "TryGet Incompatible cache entry %A" x
-                )
+                    | x -> failwithf "TryGet Incompatible cache entry %A" x }
--- a/src/Equinox.Core/Cache.fs
+++ b/src/Equinox.Core/Cache.fs
@@ -1,0 +1,25 @@
+﻿namespace Equinox.Core
+
+open System
+
+type CacheItemOptions =
+    | AbsoluteExpiration of DateTimeOffset
+    | RelativeExpiration of TimeSpan
+
+[<AllowNullLiteral>]
+type CacheEntry<'state>(initialToken: StreamToken, initialState: 'state, supersedes: StreamToken -> StreamToken -> bool) =
+    let mutable currentToken, currentState = initialToken, initialState
+    member __.UpdateIfNewer(other: CacheEntry<'state>) =
+        lock __ <| fun () ->
+            let otherToken, otherState = other.Value
+            if otherToken |> supersedes currentToken then
+                currentToken <- otherToken
+                currentState <- otherState
+    member __.Value: StreamToken * 'state =
+        lock __ <| fun () ->
+            currentToken, currentState
+
+
+type ICache =
+    abstract member UpdateIfNewer: cacheItemOptions:CacheItemOptions -> key: string -> CacheEntry<'state> -> Async<unit>
+    abstract member TryGet: key: string -> Async<(StreamToken * 'state) option>
--- a/src/Equinox.Core/Equinox.Core.fsproj
+++ b/src/Equinox.Core/Equinox.Core.fsproj
@@ -29,7 +29,7 @@
 
     <PackageReference Include="FSharp.Core" Version="3.1.2.5" Condition=" '$(TargetFramework)' != 'netstandard2.0' " />
     <PackageReference Include="FSharp.Core" Version="4.3.4" Condition=" '$(TargetFramework)' == 'netstandard2.0' " />
-    <PackageReference Include="System.Runtime.Caching" Version="4.6.0" />
+    <PackageReference Include="System.Runtime.Caching" Version="4.6.0"  Condition=" '$(TargetFramework)' == 'netstandard2.0' "/>
   </ItemGroup>
 
 </Project>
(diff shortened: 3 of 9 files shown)


bartelink:
Thanks for doing this work; from first glance it looks pretty complete.

Quick thoughts:
- It comes at a good point as #62 is in progress and is about to add another store that can definitely benefit from a cache. @rajivhost
- I'm thinking I'd like to consider overlaps with #157 and #61 before releasing (might take a run at #157 in a bit and we can see if the outcome of that helps in any way) 
- I'd definitely like a prefix on the names of the Cache impl packages - i.e. Equinox.Cache.SystemRuntime and Equinox.Caching.MicrosoftExtensions or similar naming style in alignment with how FsCodec name plugins
- 🤔 I guess we have integration tests but I'm wondering about whether there's a good way to do unit tests
- 🤔 maybe the cache interface should go somewhere other than in the core Equinox assembly (there are some StoreHelpers that are inlined into each store impl which maybe should go into an Equinox.Storage package) - the fact that Equinox.dll has 240 lines of fs including comments is important (at least to me) - not saying it needs to happen as part of this; ultimately this makes the codebase more intelligible and we can likely best refactor from there onwards
- 🤔 I'm still slightly concerned about what something other than SRC actually buys anyone as noted in #145

Probably won't get to take a deeper look today but hopefully shortly

bartelink:
Some follow-on points:
- having a clear interface (especially if some unit tests and/or docs can use it)
- extracting the impls from the store impl files is a win
- it's not a major problem to introduce a new package in the main repo from a CI/CD point of view

despite suggesting introducing new packages, I'm keen to
- reduce the number of packages that need to be documented and/or considered by someone doing a 101 sample
- default to embedding/hiding stuff that's not something that most folks need to worry about
- not have single user features - i.e. a Microsoft.Extensions.Caching impl would need to be something everyone needs to justify putting it in the repo (right now I can't see a reason for anyone other than you to use it; until we have proof to the contrary I'm thinking that's a good way to look at it)
- while I know it's correct from many perspectives, a tree of packages or folders as big as https://github.com/eventflow/EventFlow/tree/develop/Source is a nightmare for me - being able to simply clone `EventStore.fs` in an application to hack around a bug or add a some logging is for me an important feature of having essentially 3 files of consequence in Equinox as a whole if you get my drift

This is making me think the best approach should be along the following lines
1. extract the interface and provide the seam - probably initially into the Equinox DLL
2. you can validate by having your caching impl in a separate assembly, satisfying yourself it works
3. initially we leave the SRC impl in each store for now
4. at such time as a stronger need for an Equinox.Storage DLL presents itself, move ICache and the default impl there
5. iff there's a reason to have more that one new DLL, consider having the SRC cache impl in its own assembly
6. if there was a Equinox.Cache.SRC and >2 users of Equinox.Cache.MEC, then hosting it in the main repo would start to make sense

For now, I'd be comfortable with working step 1, but prior to taking most of the other steps, I'd be looking to:
- cut a final v2
- round out stuff like  #157 #61 
- the other consideration is that if we're to have a #62 impl arriving soon, that's likely to force and/or guide some  of the issues in steps 2-6 above

Does any of this seem reasonable to you?

bartelink:
I'm looking to implement #157 soon - ideally that'd be after this gets rebased and/or cherrypicked onto master now #164 is complete.
@DSilence do you see any time opening up to do that your side any time soon? Centralizing the SRC impl would make a great addition for a V2 release (esp if/when SqlStreamStreamStore support arrives), and hopefully shouldn't be _too much_ work given you've done the hard bit ;)

DSilence:
Agree with the points above.
While I would personally prefer extra modularity in assemblies, given that you want to keep things simple for now I'm totally fine to go with your suggestions.

I've updated the PR and rebased it on the latest master. The interface (and related base types) were moved to Core package (to a separate Cache.fs file), and implementations are made part of EventStore.fs and Cosmos.fs. My only issue with that is that those packages remain dependent on System.Runtime.Caching - even if another caching library is used, but it's something I can live with 😄.
The Microsoft.Extensions.Caching integration package was removed for now - I might end up publishing it as a separate package (or make a blog post with the implementation sample) if you don't mind.

bartelink:
Thanks! Given it wouldn't materially affect ones dependency tree, what would you think about moving to a single impl of Cache as `Equinox.Cache` but housed in the `Equinox.Core` package (one thing prompting this is that we're about to sprout SqlStreamStore support, another is that i can simplify the wiring in `eqxweb`) ?
Blog posts, articles, edits to DOCUMENTATION.md are all more than welcome any time of course ;)

DSilence:
All up for it. I'll update the pull request.

DSilence:
Done, please take a look.

samples/Infrastructure/Storage.fs
```diff
@@ -76,7 +76,7 @@ module Cosmos =
         let conn = connector.Connect("equinox-tool", discovery) |> Async.RunSynchronously
         let cacheStrategy =
             if cache then
-                let c = Caching.Cache("equinox-tool", sizeMb = 50)
+                let c = Equinox.Cache("equinox-tool", sizeMb = 50)
```
bartelink: now this is more general, the sample can centralize the creation of the cache. Doesn't have to happen in this PR though

src/Equinox.Core/Cache.fs
```diff
@@ -0,0 +1,58 @@
...
+        lock __ <| fun () ->
+            currentToken, currentState
+
+
+type ICache =
+    abstract member UpdateIfNewer: cacheItemOptions:CacheItemOptions -> key: string -> CacheEntry<'state> -> Async<unit>
+    abstract member TryGet: key: string -> Async<(StreamToken * 'state) option>
+
+namespace Equinox
+open System.Runtime.Caching
```
bartelink: I like it in one file as you've done. blank lines around the opens

src/Equinox.Core/Cache.fs
```diff
@@ -0,0 +1,58 @@
...
+            | AbsoluteExpiration absolute -> new CacheItemPolicy(AbsoluteExpiration = absolute)
+            | RelativeExpiration relative -> new CacheItemPolicy(SlidingExpiration = relative)
+
+        interface ICache with
+
+            member this.UpdateIfNewer cacheItemOptions key entry =
+                let policy = getPolicy cacheItemOptions
+                match cache.AddOrGetExisting(key, box entry, policy) with
+                | null ->
+                    async.Return ()
```
bartelink: wrap the whole method in `async {` instead unless you're sure there's a significant perf benefit

src/Equinox.Core/Equinox.Core.fsproj
```diff
@@ -28,6 +29,7 @@
 
     <PackageReference Include="FSharp.Core" Version="3.1.2.5" Condition=" '$(TargetFramework)' != 'netstandard2.0' " />
     <PackageReference Include="FSharp.Core" Version="4.3.4" Condition=" '$(TargetFramework)' == 'netstandard2.0' " />
+    <PackageReference Include="System.Runtime.Caching" Version="4.6.0" />
```
bartelink: need to also have the condition too - copy from the stores and remove from them

src/Equinox.Cosmos/Cosmos.fs
```diff
@@ -874,62 +875,39 @@ type private Category<'event, 'state>(gateway : Gateway, codec : IUnionEncoder<'
...
                 interceptAsync (inner.Load containerStream log) (snd containerStream)
             member __.TrySync (log : ILogger) (Token.Unpack (_container,stream,_) as streamToken,state) (events : 'event list)
                 : Async<SyncResult<'state>> = async {
                 let! syncRes = inner.TrySync log (streamToken, state) events
                 match syncRes with
-                | SyncResult.Conflict resync ->         return SyncResult.Conflict (interceptAsync resync stream)
-                | SyncResult.Written (token', state') ->return SyncResult.Written (intercept stream (token', state')) }
+                | SyncResult.Conflict resync -> return SyncResult.Conflict(interceptAsync resync stream)
+                | SyncResult.Written(token', state')
+                     ->
```
bartelink: inconsistent indentation?

src/Equinox.Cosmos/Cosmos.fs
```diff
@@ -775,6 +775,7 @@ open Microsoft.Azure.Documents
 open Serilog
 open System
 open System.Collections.Concurrent
+open System.Runtime.Caching
```
bartelink: is this still needed?

src/Equinox.Cosmos/Cosmos.fs
```diff
@@ -939,16 +917,18 @@ type private Folder<'event, 'state>
...
+                let! batched = category.Load inspectUnfolds containerStream fold initial isOrigin log
+                let cached tokenAndState = category.LoadFromToken tokenAndState fold isOrigin log
+                match readCache with
+                | None -> return batched
+                | Some (cache : ICache, prefix : string) ->
+                    let! cacheItem = cache.TryGet(prefix + snd containerStream)
+                    match cacheItem with
+                    | None -> return batched
+                    | Some tokenAndState -> return! cached tokenAndState
+            }
```
bartelink: In this file, I put the brace on prev line for conservation of vertical space (you don't have to, but don't be surprised if I move it up later!)

src/Equinox.EventStore/EventStore.fs
```diff
@@ -442,44 +442,16 @@ type private Category<'event, 'state>(context : Context, codec : FsCodec.IUnionE
...
-            | :? CacheEntry<'state> as existingEntry -> Some existingEntry.Value
-            | x -> failwithf "TryGet Incompatible cache entry %A" x
-
     /// Forwards all state changes in all streams of an ICategory to a `tee` function
-    type CategoryTee<'event, 'state>(inner: ICategory<'event, 'state, string>, tee : string -> StreamToken * 'state -> unit) =
+    type CategoryTee<'event, 'state>(inner: ICategory<'event, 'state, string>, tee : string -> StreamToken * 'state -> Async<unit>) =
         let intercept streamName tokenAndState =
-            tee streamName tokenAndState
-            tokenAndState
+            async{
```
bartelink: up to prev line (same for last brace)

src/Equinox.EventStore/EventStore.fs
```diff
@@ -489,26 +461,31 @@ module Caching =
...
     let applyCacheUpdatesWithSlidingExpiration
-            (cache: Cache)
+            (cache: ICache)
             (prefix: string)
             (slidingExpiration : TimeSpan)
             (category: ICategory<'event, 'state, string>)
             : ICategory<'event, 'state, string> =
-        let policy = new CacheItemPolicy(SlidingExpiration = slidingExpiration)
-        let addOrUpdateSlidingExpirationCacheEntry streamName = CacheEntry >> cache.UpdateIfNewer policy (prefix + streamName)
+        let cacheEntryGenerator (initialToken: StreamToken, initialState: 'state) = new CacheEntry<'state>(initialToken, initialState, Token.supersedes)
```
bartelink: perhaps mkCacheEntry or create* to align with other naming?

src/Equinox.EventStore/EventStore.fs
```diff
@@ -489,26 +461,31 @@ module Caching =
...
     let loadAlgorithm streamName initial log =
-        let batched = category.Load fold initial streamName log
-        let cached token state = category.LoadFromToken fold state streamName token log
-        match readCache with
-        | None -> batched
-        | Some (cache : Caching.Cache, prefix : string) ->
-            match cache.TryGet(prefix + streamName) with
-            | None -> batched
-            | Some (token, state) -> cached token state
+        async {
```
bartelink: up and unindent

bartelink:
Looks great. Some minor picky comments regarding my/house style adherence added.

src/Equinox.Core/Cache.fs
DSilence: Perf hit almost nonexistant. Will wrap in `async` for better readability.

samples/Infrastructure/Storage.fs
DSilence: slightly scared to touch it in scope of this PR. Will leave it as is for now if you don't mind.

bartelink: that's reasonable - small PRs are good PRs too

src/Equinox.Cosmos/Cosmos.fs
```diff
@@ -874,62 +874,37 @@ type private Category<'event, 'state>(gateway : Gateway, codec : IUnionEncoder<'
...
-            | :? CacheEntry<'state> as existingEntry -> Some existingEntry.Value
-            | x -> failwithf "TryGet Incompatible cache entry %A" x
-
     /// Forwards all state changes in all streams of an ICategory to a `tee` function
-    type CategoryTee<'event, 'state>(inner: ICategory<'event, 'state, Container*string>, tee : string -> StreamToken * 'state -> unit) =
+    type CategoryTee<'event, 'state>(inner: ICategory<'event, 'state, Container*string>, tee : string -> StreamToken * 'state -> Async<unit>) =
         let intercept streamName tokenAndState =
-            tee streamName tokenAndState
-            tokenAndState
+            async{
```
bartelink: collapse `async {` to prev line and `}` up a line to regain 2 lines ;)

src/Equinox.Cosmos/Cosmos.fs
```diff
@@ -939,15 +912,15 @@ type private Folder<'event, 'state>
         ?readCache) =
     let inspectUnfolds = match mapUnfolds with Choice1Of3 () -> false | _ -> true
     interface ICategory<'event, 'state, Container*string> with
-        member __.Load containerStream (log : ILogger): Async<StreamToken * 'state> =
-            let batched = category.Load inspectUnfolds containerStream fold initial isOrigin log
+        member __.Load containerStream (log : ILogger): Async<StreamToken * 'state>  = async {
+            let! batched = category.Load inspectUnfolds containerStream fold initial isOrigin log
```
bartelink: this is a behaviour change - should be return! on L922 - same in other callsite

src/Equinox.Core/Equinox.Core.fsproj
```diff
@@ -28,6 +29,8 @@
 
     <PackageReference Include="FSharp.Core" Version="3.1.2.5" Condition=" '$(TargetFramework)' != 'netstandard2.0' " />
     <PackageReference Include="FSharp.Core" Version="4.3.4" Condition=" '$(TargetFramework)' == 'netstandard2.0' " />
+    <PackageReference Include="System.Runtime.Caching" Version="4.5.0"  Condition=" '$(TargetFramework)' == 'netstandard2.0' "/>
```
bartelink: pretty sure these can now be removed elsewhere as they're transitively referenced

src/Equinox.Cosmos/Cosmos.fs
DSilence: Totally missed that 🤦‍♂. Looks like this also caused an integration test failure.

bartelink: Yeah, I've been 🙈about automating the running of those for too long; Unfortunately it's not in danger of getting to the top of my TODO list for a while :(

bartelink:
🎉Looks perfect now; thanks for bearing with me over the course of all the iterations.
Aside from opening up the extension point, the forcing of the introduction of Equinox.Core and moving the cache impls to a central location really improves the legibility and hence approachability of the store implementation files. Merging now cc
 @rajivhost

PR #163, opened 2019-09-21
Change Target.AggregateIdEmpty to ResolveOption

bartelink:
- [x] Change `AggregateIdEmpty` to `ResolveOption.AssumeEmpty`
This paves the way for using that as a way to request a `UseCached` re #157


bartelink:
Deferred task to introduce a `ResolveOption.UseCached` re #157
It appears it might be best to approach this by first introducing the changes in #161 as a basis

PR #170, opened 2019-10-16
Optional `skipSslCheck` flag to Equinox.Cosmos.Connector

Kelvin4702:
The CosmosDB emulator uses a self-signed certificate that is issued only to `localhost` by default. In order to use the emulator from a Docker container, one must either
1) generate a new certificate with the respective alternate subject names, export it, then install it into the Docker image,
2) or override the default handler in the DocumentClient to skip SSL verification

This change provides a flag for using option 2.

Context: https://docs.microsoft.com/en-us/azure/cosmos-db/local-emulator#running-on-mac-or-linux
This might no longer be required [if/when there is a Linux Docker image available for the emulator](https://feedback.azure.com/forums/263030-azure-cosmos-db/suggestions/18533509-add-documentdb-emulator-support-for-mac-os-x-and).


src/Equinox.Cosmos/Equinox.Cosmos.fsproj
```diff
@@ -27,8 +27,8 @@
 
     <PackageReference Include="FsCodec.NewtonsoftJson" Version="1.0.0-pr.20.rc2.5" />
     <PackageReference Include="FSharp.Control.AsyncSeq" Version="2.0.21" />
-    <PackageReference Include="Microsoft.Azure.DocumentDB" Version="2.0.0" Condition=" '$(TargetFramework)' != 'netstandard2.0' " />
-    <PackageReference Include="Microsoft.Azure.DocumentDB.Core" Version="2.0.0" Condition=" '$(TargetFramework)' == 'netstandard2.0' " />
+    <PackageReference Include="Microsoft.Azure.DocumentDB" Version="2.7.0" Condition=" '$(TargetFramework)' != 'netstandard2.0' " />
```
bartelink: putting a hard 2.7 dependency in is pretty debatable for me; in general I try to be conservative and let apps do the upgrade as desired. Will hence seek to find the lowest that compiles (i.e probably 2.2.0 as you suggest) - will double check this is tenable for existing uses too

src/Equinox.Cosmos/Cosmos.fs
```diff
@@ -1080,7 +1080,9 @@ type Connector
         [<O; D(null)>]?writeRetryPolicy,
         /// Additional strings identifying the context of this connection; should provide enough context to disambiguate all potential connections to a cluster
         /// NB as this will enter server and client logs, it should not contain sensitive information
-        [<O; D(null)>]?tags : (string*string) seq) =
+        [<O; D(null)>]?tags : (string*string) seq,
+        /// Skips SSL verification when set to true, i.e. for working with the CosmosDB Emulator (default false)
+        [<O; D(null)>]?skipSslCheck : bool) =
```
bartelink: Suggested name: `bypassCertificateValidation` (avoiding SSL/TLS debates) ?

src/Equinox.Cosmos/Cosmos.fs
```diff
@@ -1107,7 +1109,13 @@ type Connector
                 yield name
                 match tags with None -> () | Some tags -> for key, value in tags do yield sprintf "%s=%s" key value }
             let sanitizedName = name.Replace('\'','_').Replace(':','_') // sic; Align with logging for ES Adapter
-            let client = new Client.DocumentClient(uri, key, clientOptions, Nullable(defaultArg defaultConsistencyLevel ConsistencyLevel.Session))
+            let client =
```
bartelink:
```
let consistencyLevel = Nullable(defaultArg defaultConsistencyLevel ConsistencyLevel.Session)
let client =
    if defaultArg bypassCertificateValidation false then
        let inhibitCertCheck = new System.Net.Http.HttpClientHandler(ServerCertificateCustomValidationCallback = fun _ _ _ _ -> true)
        new Client.DocumentClient(uri, key, inhibitCertCheck, clientOptions, consistencyLevel)
    else new Client.DocumentClient(uri, key, clientOptions, consistencyLevel)
```

bartelink:
@enricosada this PR proposes to up the DocDb client ver to 2.2.0 (from 2.0.0)
Equinox once had a 2.2.x dependency (in ~ 1.0.2) - is upping the requirement a significant problem warranting publishing of this as a beta package or do we feel overriding to 2.2 is The Right Thing To Do?

bartelink:
Thanks! As agreed out of band, I'll apply the tweaks above separately

PR #171, opened 2019-10-19
Add Cosmos.Connector.CreateClient

bartelink:
Addendum to #170, exposing a `CreateClient` which exposes the `DocumentClient`
Intended for use with https://github.com/jet/propulsion/pull/40
@Kelvin4702

### DIFF
--- a/src/Equinox.Cosmos/Cosmos.fs
+++ b/src/Equinox.Cosmos/Cosmos.fs
@@ -1075,7 +1075,8 @@ type Connector
         /// Maximum number of times attempt when failure reason is a 429 from CosmosDb, signifying RU limits have been breached
         maxRetryAttemptsOnRateLimitedRequests: int,
         /// Maximum number of seconds to wait (especially if a higher wait delay is suggested by CosmosDb in the 429 response)
-        maxRetryWaitTimeOnRateLimitedRequestsSeconds: int,
+        // naming matches SDK ver >=3
+        maxRetryWaitTimeOnRateLimitedRequests: TimeSpan,
         /// Log to emit connection messages to
         log : ILogger,
         /// Connection limit for Gateway Mode (default 1000)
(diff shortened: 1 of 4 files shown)


src/Equinox.Cosmos/Cosmos.fs
```diff
@@ -1110,35 +1110,43 @@ type Connector
...
-        match discovery with Discovery.UriAndKey(databaseUri=uri; key=key) -> connect (uri,key)
+    let logName (uri : Uri) name =
+        let name = String.concat ";" <| seq {
+            yield name
+            match tags with None -> () | Some tags -> for key, value in tags do yield sprintf "%s=%s" key value }
+        let sanitizedName = name.Replace('\'','_').Replace(':','_') // sic; Align with logging for ES Adapter
+        log.ForContext("Uri", uri).Information("CosmosDb Connection Name {connectionName}", sanitizedName)
+    let mkClient name skipLog = function
+        | Discovery.UriAndKey(databaseUri=uri; key=key) ->
+            if not skipLog then logName uri name
```
bartelink: 🤔 the name should probably also be fed into the `DocumentClient` for inclusion in cosmosdb's analytics. I dont know how to do that - the `User-Agent` doesnt feel right for that

src/Equinox.Cosmos/Cosmos.fs
```diff
@@ -1109,36 +1109,40 @@ type Connector
...
-                    new Client.DocumentClient(uri, key, inhibitCertCheck, clientOptions, consistencyLevel) // overload introduced in 2.2.0 SDK
-                else new Client.DocumentClient(uri, key, clientOptions, consistencyLevel)
-            log.ForContext("Uri", uri).Information("CosmosDb Connection Name {connectionName}", sanitizedName)
-            do! client.OpenAsync() |> Async.AwaitTaskCorrect
-            return client }
-
-        match discovery with Discovery.UriAndKey(databaseUri=uri; key=key) -> connect (uri,key)
-
-    /// ClientOptions (ConnectionPolicy with v2 SDK) for this Connector as configured
-    member __.ClientOptions : Client.ConnectionPolicy = clientOptions
```
bartelink: This is breaking, but on the plus side, it will force folks to upgrade Propulsion.Cosmos libs and makes this API clearer (it's once again about connecting)

bartelink: got cold feet - its there 'just in case'

src/Equinox.Cosmos/Cosmos.fs
bartelink: someone'll bring it some day ;)

PR #177, opened 2019-11-07
Add eqx dump cosmos

bartelink:
Adds an `eqx dump` command that can be used to inspect the raw [json] data of a specified stream from the console.

- [x] This initial implementation is only wired for `Equinox.Cosmos` (the unique need here is to easily render the compressed `u`nfolds field)

image
image

- `-vc` will trigger emitting storage logging
- `-v` adds detail to said logging

- [x] handle any store type


bartelink:
example output from `eqx run es`-generated-data
image

image

PR #178, opened 2019-11-13
AccessStrategy naming cleanup; add RollingState strategy

jgardella:
WIP renames done with @bartelink, open to suggestions.

- Renames several Cosmos access strategies
- Makes Cosmos access strategy non-optional (`None` case now represented by `AccessStrategy.Unoptimized`)
- New Cosmos access strategy `RollingState`


src/Equinox.Cosmos/Cosmos.fs
```diff
@@ -998,30 +998,36 @@ type CachingStrategy =
...
     /// Provides equivalent performance to Unfolded, just simplified function signatures
-    | Snapshot of isValid: ('event -> bool) * generate: ('state -> 'event)
-    /// Treat every known event type in the Union as being an Origin event
-    | AnyKnownEventType
+    | Snapshot of isOrigin: ('event -> bool) * toSnapshot: ('state -> 'event)
+    /// Instead of actually storing the events representing the decisions, only ever update the snapshot stored in the tip
+    | RollingState of toSnapshot: ('state -> 'event)
+    /// Allow events that pass the `isOrigin` test to be used in lieu of folding all the events from the start of the stream
+    /// When saving, uses `toEvents` to 'unfold' the 'state, representing it as event records to be stored in the Tip with efficient read cost
+    | CustomUnfold of isOrigin: ('event -> bool) * toEvents: ('state -> 'event seq)
```
jgardella: `MultiSnapshot` and `toEvents` -> `toSnapshots`

src/Equinox.Cosmos/Cosmos.fs
```diff
@@ -998,30 +998,36 @@ type CachingStrategy =
...
     /// Allow produced events to be transmuted to unfolds.
     /// In this mode, Optimistic Concurrency Control OCC is based on the _etag (rather than the normal Expected Version strategy)
-    | RollingUnfolds of isOrigin: ('event -> bool) * transmute: ('event list -> 'state -> 'event list*'event list)
+    | Custom of isOrigin: ('event -> bool) * transmute: ('event list -> 'state -> 'event list*'event list)
+    /// Treat every known event type in the Union as being an Origin event
+    | EventsAreState
```
jgardella: `LatestKnownEvent`

jgardella: And EventStore one too.

bartelink: Ditto SqlStreamStore

bartelink:
Need to update documentation.md, readme.md and changelog too...

src/Equinox.EventStore/EventStore.fs
```diff
@@ -396,7 +396,7 @@ type Context(conn : Connection, batching : BatchingPolicy) =
 
 [<NoComparison; NoEquality; RequireQualifiedAccess>]
 type AccessStrategy<'event,'state> =
-    | EventsAreState
+    | LatestKnownEvent
     | RollingSnapshots of isValid: ('event -> bool) * compact: ('state -> 'event)
```
jgardella: Any renames here for consistency with Cosmos changes? @bartelink

bartelink: isValid and compact should also become `isOrigin` and `toSnapshot`

RollingSnapshots should say:

```
/// Ensures a snapshot is always present (embedded in the stream as an event) every `batchSize` events using the supplied `toSnapshot` function to generate an event from which the state can be reconstituted upon decoding. See https://eventstore.org/docs/event-sourcing-basics/rolling-snapshots/index.html
```

.. or something better

src/Equinox.SqlStreamStore/SqlStreamStore.fs
```diff
@@ -405,8 +405,10 @@ type Context(conn : Connection, batching : BatchingPolicy) =
 
 [<NoComparison; NoEquality; RequireQualifiedAccess>]
 type AccessStrategy<'event,'state> =
-    | EventsAreState
-    | RollingSnapshots of isValid: ('event -> bool) * compact: ('state -> 'event)
+    | LatestKnownEvent
+    /// Ensures a snapshot is always present (embedded in the stream as an event) every `batchSize` events using
```
jgardella: Anything else to call out in this comment? @bartelink

bartelink: how about including the link to the ES article?
also should clone the comment for LatestKnownEvent

jgardella: For the `LastestKnownEvent` comment:

`Treat every known event type in the Union as being an Origin event` -> `Load latest event type which is defined in the events union`?

src/Equinox.Cosmos/Cosmos.fs
```diff
@@ -998,30 +998,36 @@ type CachingStrategy =
...
+    | RollingState of toSnapshot: ('state -> 'event)
+    /// Allow events that pass the `isOrigin` test to be used in lieu of folding all the events from the start of the stream
+    /// When saving, uses `toEvents` to 'unfold' the 'state, representing it as event records to be stored in the Tip with efficient read cost
+    | MultiSnapshot of isOrigin: ('event -> bool) * toSnapshots: ('state -> 'event seq)
     /// Allow produced events to be transmuted to unfolds.
     /// In this mode, Optimistic Concurrency Control OCC is based on the _etag (rather than the normal Expected Version strategy)
-    | RollingUnfolds of isOrigin: ('event -> bool) * transmute: ('event list -> 'state -> 'event list*'event list)
+    | Custom of isOrigin: ('event -> bool) * transmute: ('event list -> 'state -> 'event list*'event list)
+    /// Treat every known event type in the Union as being an Origin event
+    | LatestKnownEvent
```
bartelink: xmldoc
```
/// Treat every event that we can decode as an Origin event, i.e. only ever load one event and build state from that, implying we never need to reference any events aside from the latest to know the state
/// <remarks>
Used where every event represented on the stream is sufficient to build a state from
/// A copy of the event is also retained in the `Tip` document in order that the state of the stream can be retrieved using a single (cached, etag-checked) point read.
/// </remarks>
```

the reference to the Tip document line should not go on the ES or SSS versions but the rest should be the same

src/Equinox.SqlStreamStore/SqlStreamStore.fs
bartelink: (added a separate comment - feel free to improve on it. I was reluctant to use Union as it doesn't strictly have to be a Union, i.e. if not using the FsCodec.NewtonsoftJson.Codec but I do accept its clearer to just talk about the common case)

PR #180, opened 2019-11-14
- fixed compilation and run of counter example; - made comments more …

dzmitry-lahoda:
…consistent


dzmitry-lahoda:
i have used Rider to run code, no VS

bartelink:
Thanks; it's been a while since I've alt-entered that one!

PR #186, opened 2020-01-16
Fix #185: Add EventStore Connector missing configuration for ClientConnectionTimeout and GossipTimeout

### LINKED ISSUE #185: Equinox.EventStore.Connector is missing some configuration properties
`Equinox.EventStore.Connector` contains configuration not for all properties of `ConnectionSettingsBuilder`, for example there are no way to set `ClientConnectionTimeout` and `GossipTimeout`. That makes problems when connection time to some nodes and clusters exceeds default values (1000 ms for both settings)

AndrewRublyov:
Resolves #185, see issue description

### DIFF
--- a/src/Equinox.EventStore/EventStore.fs
+++ b/src/Equinox.EventStore/EventStore.fs
@@ -623,7 +623,8 @@ type ConnectionStrategy =
 type Connector
     (   username, password, reqTimeout: TimeSpan, reqRetries: int,
         [<O; D(null)>]?log : Logger, [<O; D(null)>]?heartbeatTimeout: TimeSpan, [<O; D(null)>]?concurrentOperationsLimit,
-        [<O; D(null)>]?readRetryPolicy, [<O; D(null)>]?writeRetryPolicy,
+        [<O; D(null)>]?readRetryPolicy, [<O; D(null)>]?writeRetryPolicy, [<O; D(null)>]?clientConnectionTimeout,
+        [<O; D(null)>]?gossipTimeout,
         /// Additional strings identifying the context of this connection; should provide enough context to disambiguate all potential connections to a cluster
         /// NB as this will enter server and client logs, it should not contain sensitive information
         [<O; D(null)>]?tags : (string*string) seq) =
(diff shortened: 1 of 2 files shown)


src/Equinox.EventStore/EventStore.fs
```diff
@@ -623,7 +623,8 @@ type ConnectionStrategy =
 type Connector
     (   username, password, reqTimeout: TimeSpan, reqRetries: int,
         [<O; D(null)>]?log : Logger, [<O; D(null)>]?heartbeatTimeout: TimeSpan, [<O; D(null)>]?concurrentOperationsLimit,
-        [<O; D(null)>]?readRetryPolicy, [<O; D(null)>]?writeRetryPolicy,
+        [<O; D(null)>]?readRetryPolicy, [<O; D(null)>]?writeRetryPolicy, [<O; D(null)>]?clientConnectionTimeout,
```
bartelink: can you add a line break after writeRetryPolicy please?
prob should be on same line as gossipTimeout ?
as the gossip happens before the connect maybe also swap the order (ditto order of application below)

AndrewRublyov: I don't think that order in that case really matter, but I'll change it

bartelink: on reflection, given gossip timeout is only relevant for gossip, the order of the two args probably makes sense as it is
(also arguably the gossip options could or should go into the Discovery union, but thats probably a bridge too far given the next step is to mint a `Equinox.EventStore.Grpc` for ES v6 ;) )

AndrewRublyov: Changed

bartelink: Thanks, will merge (unless you want to write the changelog.md entry first...)

AndrewRublyov: Updated `CHANGELOG.md`

bartelink: Wonderful, thanks!, merging shortly

Issue #188, opened 2020-01-30
Several connections in one environment variable


dzmitry-lahoda:
Issue:
- we have runtime when we need several connections to multiple event stores
- it is impossible to configure via JSON

Solution:
- scheme as in next test

Otherusages:
- one liner connection string from notebook to make streams research more interactive

```fsharp
module Tests
open Equinox.EventStore
open Expecto
open System

type Connector with
    static member Connect (uri:Uri, connectionName) =
        // TODO: drop proper exception
        // TODO: consider what if no : presented
        // TODO: consider password contains non url encoded chracters
        // TODO: consider ?requestTimeoutMilliseconds= and requestRetries=
        let namePassword = uri.UserInfo.Split(":")
        let discovery =
          match uri.Scheme with
          | "uri" ->  uri.Host |> Uri |> Discovery.Uri
          | "gossipdns" -> uri.Host |> Discovery.GossipDns   
        let connector = Connector(namePassword.[0], namePassword.[1], TimeSpan.FromSeconds 5., 10)
        connector.Connect(connectionName, discovery) 



[<Tests>]
let tests =
  testList "samples" [
    let url = "gossipdns://admin:changeit@localhost:1234?requestTimeoutMilliseconds=5000&requestRetries=10"

    test "format" {
      let parsed = Uri(url)
      
      Expect.equal parsed.Scheme "gossipdns" "Scheme"
      Expect.equal parsed.UserInfo "admin:changeit" "UserPassword"
      Expect.equal parsed.Host "localhost" "Host"      
    }

    testAsync "factory1" {
      let parsed = Uri(url)
      let! connection = Connector.Connect(parsed,"Test")
      Expect.isNotNull connection "Created"
    }    
  ]

```

dzmitry-lahoda:
Feel free to close idea if does not suit. Thanks

bartelink:
1) re connection strings and/or easily serializable forms of ES conn strs

Have you seen the intrinsic connection string support in the ES client?
it might be possible to wire that in as another DU case and/or overload on the `Connect` method
but, really we're talking an alternate way of making an IEventStoreConnection I guess - if it needs to parallel Connector, which is OK to do

[...]

PR #196, opened 2020-02-17
Add Equinox.EventStoreDb - gRPC support

bartelink:
This introduces a port of the `Equinox.EventStore` module entitled `Equinox.EventStore.Grpc` which:
- uses `EventStore.Client.Grpc.Streams` v `22.0.0`

In doing this, the `Equinox.EventStore` package remains but:
- the Sample's integration tests shift to use the gRPC package instead
- is intended for use against server versions `21.10` and later, i.e. there is no logic to use V5 server compatibility mode etc 
- references `EventStore.Client` v `22.0.0-preview`

As part of the above, the guidance changes from running a local ESDB (v5) instance obtained via  Chocolatey (on Windows) or `brew` (on MacOS) to using a 3 node cluster running under `docker compose`

related: https://github.com/jet/equinox/issues/188


bartelink:
This module should be renamed to `Equinox.EventStoreDb` (to disambiguate from `Equinox.EventStore`), which will continue to depend on `EventStore.Client`, (and align with EventStore's new consistent naming).

bartelink:
Would love to see this implemented as it's abundantly clear that the gRPC interface is the only game in town going forward for EventStoreDB and simply makes sense to migrate to.

I don't have time to do this at present (and my current team I'm on are still on V5).

dharmaturtle:
Hi! What would it take to complete this PR? When I [diff `EventSToreGrpc.fs` with the main `EventStore.fs`](https://www.diffchecker.com/zwsAhI2A), it isn't clear to me what's missing (other than the one TODO). Is it just updating the client to the LTS (`20.10`) and resolving the conflicts?

bartelink:
> Edit: Ah - it doesn't compile.

Unfortunately yes, it's very incomplete @dharmaturtle :( 

I was making a first stab at it during a thing I believe we used to call "a conference", prior to the 20.x release :D

Since this point, the ESDB Folk have added some good getting started and porting guides. I'd also ask for advice in `#eventstore` in the DDD-CQRS-ES slack

dharmaturtle:
Do we want to keep the `Discovery` and `NodePreference` APIs? I'm not sure why we're abstracting away `EventStoreClientSettings`. Would it be reasonable to have `Connector`'s `Connect` and `Establish` take an `EventStoreClientSettings`, and let the user deal with the construction of the settings?

If we really want, we could build a fluent `Settings` builder, but I don't really see it as the job of this library to hide ESDB from the client. This would also make for less code to maintain, and we won't have to add a new param for each possible configuration option.

bartelink:
High level principles of V3 are noted in https://github.com/jet/equinox/issues/232. If you feel there's anything below in conflict with that, we should address it by editing the issue.

>Do we want to keep the Discovery and NodePreference APIs? I'm not sure why we're abstracting away EventStoreClientSettings. Would it be reasonable to have Connector's Connect and Establish take an EventStoreClientSettings, and let the user deal with the construction of the settings?

It makes sense to start from your baseline use cases, without necessarily looking to provide 100% source compatibility with `Equinox.EventStore.Connector`. The new connection string-centric API also implies changes to the idiom. 

>If we really want, we could build a fluent Settings builder, but I don't really see it as the job of this library to hide ESDB from the user.

The fluent settings builder is something that `Connector` seeks to hide. Some key aims:
- force specification of some settings that one would rarely want to blindly leave at the default
- having a terse way to represent configurations in F# (though it has been used from C# and in general the intention is that the APIs of Equinox should be consumable from C# without it feeling unnatural)
- providing a paved path for the best default way to connect (i.e. connect to a cluster, not TCP single node 'demo mode'). The fact that the product has better defaults in >= 20.x makes a lot of this go away
- provide an implementation of the Twin connection mode to reduce the time to get in sync when a `WrongExpectedVersionException` cases arises until https://github.com/EventStore/EventStore/issues/1652 or something similar is implemented.

While I'm not totally against fluent settings builders, I default to not using them as:
- they're a lot of code, which were really trying to avoid - i.e. if we can do a Connector with some defaults and options in 50 lines or a fluent builder in 150-200 when the whole impl is 600, we're doing the Connector.

>This would also make for less code to maintain, and we won't have to add a new param for each possible configuration option.

This is the key bit. Im happy for you to chuck it all out and do something reasonable. If anyone has a need for extension, the normal approaches apply
- copy the `type Connector` locally, rename it and use that (this is a fundamental principle of all of Equinox - you should be able to work around an issue or add a line of logging in 5 minutes)
- if it makes sense, make a PR at your leisure; await a release; remove your temporary hack version
- just write the entire thing yourself and/or use some other connection logic as you say

Some other notes about this PR in general:

- Specifically wrt the notion of having a Connector as such, it should be noted that there's been some renaming/restructuring to attempt to reduce the number of concepts at top level - these were all driven from the CosmosStore side. I'll likely do some tweaks after this PR.

- Perhaps the best start is to re-copy `EventStore.fs` in order to bring in other global changes such as we go from the `/v2` branch (and 2.x packages) to what's on master and lock the API as a V3 final release.

- Because this is a new codebase (same as `Equinox.CosmosStore` vs `.Cosmos`), there's an opportunity to throw stuff out and/or simplify APIs. -> more than happy for you to venture any other changes too

## Naming / conflict

The names `Equinox.CosmosStore`/`MemoryStore`/`EventStore` are intended to
a) have a common suffix to convey the central point that they are swappable
b) enable us to to have an `Equinox.Cosmos` assembly (targeting V2 SDK) sit along a different named one that targets the V3 SDK while people are transitioning
c) same as point b for Equinox.EventStore - the grpc client has a new name/namespace so they could sit side by side - enable some code in a given process to talk TCP over one set of DLLs while new code uses the new client during transition. NOTE there was a dependency conflict in EventStore.Client vs the Grpc one such that they had conflicting protobuf dependencies that were not resolvable. _I believe this has been remedied_

Note the current name (`.Grpc`) was to align with the pre-release name of V20 - in general the intention is to mirror the naming in the underlying store unless there's a reason not to.

🤔  If it makes sense, perhaps the naming for the new packages can become
a) `Equinox.EventStoreDb` - match name of the product (except I hate `DB` casing!)
b) `Equinox.CosmosDb` - match name of the product (except I hate `DB` casing!)
c) `Equinox.MemoryStore` - maybe `TransientDb`,`DictionaryDb` to match :D

bartelink:
While not every aspect has been ported (there are some `TODO` markers in `EventStoreDb.fs` reflecting things that has not yet been ported, but have equally not been ruled out either), it's reached a point where its sufficiently code complete to go onto `master` - the rest can be completed as and when people have need, time and interest to do the work.

bartelink:
Merging this shortly - ping if you're interested in an alpha package (I still need to test it, and work on / the releasing of this is a side-effect of that, more than a goal in itself).

Thanks to @oskardudycz for assistance in getting this over the line (and not forgetting @thefringeninja's help at the very start)
cc @nodgjord in case your 👀 are not tired yet

PR #201, opened 2020-02-28
Add FAQ and detailed doc re Cosmos AccessStrategy

bartelink:
Thanks to @fnipo for the question which prompted this (and important editing suggestions), and @fluxlife whose review massively improved the legibility (it's still far from perfect, but that's on me...)


README.md
```diff
@@ -500,6 +500,75 @@ All non-alpha releases derive from tagged commits on `master`. The tag defines t
...
+The specifics of an Access Strategy depend on what makes sense for a given Store, i.e. `Equinox.Cosmos` necessarily has a significantly different set of strategies than `Equinox.EventStore` (although there is an intersection).
+
+Access Strategies only affect performance; you should still be able to infer the state of the aggregate based on the `fold` of all the `events` ever written on top of an `initial` state
+
+NOTE: its not important to select a strategy until you've actually actually modelled your aggregate, see [what if I change my access strategy](#changing-access-strategy)
+
+### `Equinox.Cosmos.AccessStrategy`
+
+TL;DR `Equinox.Cosmos`: (see also: [the storage model](DOCUMENTATION.md#Cosmos-Storage-Model) for a deep dive, and [glossary, below the table](#access-strategy-glossary) for definition of terms)
+- keeps all the events for a stream in a single single [CosmosDB _logical partition_](https://docs.microsoft.com/en-gb/azure/cosmos-db/partition-data)
```
echoluxingyu: Is "single single" to emphasize?

README.md
```diff
@@ -500,6 +500,75 @@ All non-alpha releases derive from tagged commits on `master`. The tag defines t
...
+- always has a special 'index' document (we term it the `Tip` document), per logical partition/stream that is accessible via an efficient _point read_ (it always has a CosmosDB `id` value of `"-1"`)
+- Events are stored in immutable documents in the logical partition, the `Tip` document is the only document that's ever updated.
```
echoluxingyu: never updated?

echoluxingyu: oh it is mentioned in the next line

README.md
```diff
@@ -500,6 +500,75 @@ All non-alpha releases derive from tagged commits on `master`. The tag defines t
...
+
+NOTE: its not important to select a strategy until you've actually actually modelled your aggregate, see [what if I change my access strategy](#changing-access-strategy)
+
+### `Equinox.Cosmos.AccessStrategy`
+
+TL;DR `Equinox.Cosmos`: (see also: [the storage model](DOCUMENTATION.md#Cosmos-Storage-Model) for a deep dive, and [glossary, below the table](#access-strategy-glossary) for definition of terms)
+- keeps all the events for a stream in a single single [CosmosDB _logical partition_](https://docs.microsoft.com/en-gb/azure/cosmos-db/partition-data)
+- always has a special 'index' document (we term it the `Tip` document), per logical partition/stream that is accessible via an efficient _point read_ (it always has a CosmosDB `id` value of `"-1"`)
+- Events are stored in immutable documents in the logical partition, the `Tip` document is the only document that's ever updated. 
+- all writes touch the `Tip`, so we have a natural way to invalidate any cached State for a given Stream - we retain the `_etag` of the `Tip` document, and updates (or consistent reads) are contingent on it not having changed
```
echoluxingyu: Just for my own knowledge. Does this `_etag` help with version check when writing an event? Do I need to explicitly compare the state I used for decision making and the current state when I am about the write the event to make sure they are the same before I persist anything?

echoluxingyu: Based on the next statement, looks like the Tip document doesn;t help with race condition of writes and it will just take them one by one. Is this right?

README.md
```diff
@@ -500,6 +500,75 @@ All non-alpha releases derive from tagged commits on `master`. The tag defines t
...
+  - allow one to post-process the events we are writing as required for reasons of optimization
+
+#### `Cosmos` Access Strategy overviews
+
+| Strategy | TL;DR | `Tip` document maintains | Best suited for |
+| :--- | :--- | :--- | :--- |
+| `Unoptimized` | Keep Events only (but there's still an (empty) `Tip` document) | Count of event (and ability to have any insertion invalidate the cache for any reader) | No load, event counts or event sizes worth talking about. <br/> initial implementations. |
+| `LatestKnownEvent` | Special mode for when every event is completely independent | A copy of the most recent event, together with the count | Tracking changes to a document where we're not really modelling events as such, but with optimal efficiency (every read is a point read of a single document) |
+| `Snapshot` | Keep a (single) snapshot in `Tip` at all times, guaranteed to include _all_ events | A single unfold produced by `toSnapshot` based on the `state'` (i.e., including any events being written) every time, together with the event count | Typical event-sourced usage patterns. <br/> Good default approach. |
```
echoluxingyu: When Snapshot strategy, is every writes updates the `Tip` file with the current snapshot and the event version? Based on what I have seen, I see the `Tip` get updated with the event number and `c": "PickTicketSnapshot"`, but I don't see the snapshot/state been persisted. Not sure if that is expected. If yes, my question is how does this help with a long-running stream, since it feels like it has to build the state from scratch anyway. If not, I think I am problably doing something wrong, so would potentially need some help on this.

README.md
bartelink: 😹fixing

README.md
echoluxingyu: Actually, is `d` storing the compressed state?

README.md
bartelink: That;s not a reason not to mention it - I appreciate you surfacing everything that doesn't flow for you ;)
(Added a mention on this line, I think it makes it clearer)

README.md
bartelink: the etag check comes into play for every write
writes can only be triggered by returning an event from interpret/decide
so yes, having the decide/interpret maintain enough state to make that dedup decision is an important way to reduce load, and hence RU consumption and/or contention under load (when it matters!)
And yes, the Tip document means we can quickly spot a conflict, but we resolve only be responding with the tip content and letting Transact then retry by re-running interpret/decide (hopefully if it is redundant, that will then end the cycle without a further write)
In eventstore, the EventId can be used to trigger dedup on the server - we could implement something like that in the stored proc but there are no such plans, mainly as I'm not aware of any use cases we have that dont have better solution
Do you feel this maps onto a sentence that would have helped you be certain?

bartelink: Added
> - The `interpret`/`decide` function is expected to deduplicate writes by not producing `events` if the `state` implies such updates would be redundant. Equinox does not have any internal mechanism to deduplicate events, thus having correct deduplication is the key to reducing round-trips and hence minimizing RU consumption (and the adverse effects that the retry cycles due to contention have, which will most likely arise when load is at its highest)

3 below this

README.md
bartelink: Snapshot means events and snapshots (RollingState is only updating Tip)
the snapshot itself is base64d + compressed when its in the Tip (will add a mention of that above) you can decode it with https://jgraph.github.io/drawio-tools/tools/convert.html (click decode) - maybe its not

bartelink: added a sentence in preamble

bartelink:
Thanks for the suggestions @echoluxingyu - updated; let me know if there's anything I missed

README.md
echoluxingyu: This is helpful!

README.md
bartelink: yes u.[d is compressed snapshot data

README.md
bartelink: thanks for prompting it (but others wont get to learn by working it out for themselves 😈 )

bartelink:
Going to move the overview into documentation.md as that's where two consecutive people looked for it IRL!

DOCUMENTATION.md
```diff
@@ -958,6 +958,81 @@ match res with
...
+- The `interpret`/`decide` function is expected to deduplicate writes by not producing `events` if the `state` implies such updates would be redundant. Equinox does not have any internal mechanism to deduplicate events, thus having correct deduplication is the key to reducing round-trips and hence minimizing RU consumption (and the adverse effects that the retry cycles due to contention have, which will most likely arise when load is at its highest).
+- The `unfolds` maintained in `Tip` have the bodies (the `d` and `m` fields) 1) deflated 2) base64 encoded (as everyone is reading the Tip, its worthwhile having the writer take on the burden of compressing, with the payback being that write amplification effects are reduced by readers paying less RUs to read them). The snapshots can be inspected securely via the `eqx` tool's `dump` facility, or _unsecurely_ online via the [**decode** button on this tool, _if the data is not sensitive_](https://jgraph.github.io/drawio-tools/tools/convert.html).
```
bartelink: @ylibrach NB

PR #202, opened 2020-03-05
Rebase on Codec logic extracted into FsCodec.System.Text.Json

bartelink:
Now https://github.com/jet/FsCodec/pull/38 bears the work @ylibrach prepped in #200, it's time to rebase on the packaged versions of those...


bartelink:
@ylibrach I have not actually run the tests or anything, but it compiles and uses the FsCodec alpha
I need to park it now and get onto other things so feel free to merge and/or do as you see fit with it

PR #206, opened 2020-03-18
Tip compression toggle through optional argument

ylibrach:
This PR adds an optional `compress` argument to the `Resolver`'s `Resolve` method, which defaults to `true`. Setting it to `false` will make it so that the data and metadata payloads of the `Unfold` within the `Tip` are stored in CosmosDB as uncompressed and unencoded.

### DIFF
--- a/src/Equinox.Cosmos/Cosmos.fs
+++ b/src/Equinox.Cosmos/Cosmos.fs
@@ -1140,7 +1140,7 @@ type Resolver<'event, 'state, 'context>(context : Context, codec, fold, initial,
         (   streamName : StreamName,
             [<O; D null>]?option,
             [<O; D null>]?context,
-            /// Determines whether the data and metadata payloads on the Tip document are base64 encoded and compressed; defaults to true
+            /// Determines whether the data and metadata payloads of the `u`nfolds in the Tip document are base64 encoded and compressed; defaults to true
             [<O; D true>]?compress) =
         let compress = defaultArg compress true
         match resolveTarget streamName, option with
@@ -1151,7 +1151,7 @@ type Resolver<'event, 'state, 'context>(context : Context, codec, fold, initial,
     member __.FromMemento
         (   Token.Unpack (container,stream,_pos) as streamToken,
             state,
-            /// Determines whether the data and metadata payloads on the Tip document are base64 encoded and compressed; defaults to true
+            /// Determines whether the data and metadata payloads of the `u`nfolds in the Tip document are base64 encoded and compressed; defaults to true
             [<O; D true>]?compress) =
         let skipInitialization = None
         Stream.ofMemento (streamToken,state) (resolveStream ((container,stream),skipInitialization) None None (defaultArg compress true))
--- a/src/Equinox.Cosmos/Cosmos.fs
+++ b/src/Equinox.Cosmos/Cosmos.fs
@@ -79,8 +79,19 @@ type [<NoEquality; NoComparison>] // TODO for STJ v5: All fields required unless
 type JsonCompressedBase64Converter() =
     inherit JsonConverter<JsonElement>()
 
+    static member Compress (value: JsonElement) =
+        if value.ValueKind = JsonValueKind.Null || value.ValueKind = JsonValueKind.Undefined then
+            value
+        else
+            let input = System.Text.Encoding.UTF8.GetBytes(value.GetRawText())
+            use output = new MemoryStream()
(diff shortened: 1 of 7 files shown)


src/Equinox.Cosmos/Cosmos.fs
```diff
@@ -79,8 +79,19 @@ type [<NoEquality; NoComparison>] // TODO for STJ v5: All fields required unless
...
+    static member Compress (value: JsonElement) =
+        if value.ValueKind = JsonValueKind.Null || value.ValueKind = JsonValueKind.Undefined then
+            value
+        else
+            let input = System.Text.Encoding.UTF8.GetBytes(value.GetRawText())
+            use output = new MemoryStream()
+            use compressor = new System.IO.Compression.DeflateStream(output, System.IO.Compression.CompressionLevel.Optimal)
+            compressor.Write(input, 0, input.Length)
+            compressor.Close()
+            JsonDocument.Parse("\"" + System.Convert.ToBase64String(output.ToArray()) + "\"").RootElement
```
bartelink: ouch, feel those allocations ! 
is JsonDocument.Parse definitely the most direct way to generate a JsonElement with a string body?

src/Equinox.Cosmos/Cosmos.fs
```diff
@@ -536,12 +539,22 @@ module Sync =
...
-    let mkUnfold baseIndex (unfolds: IEventData<_> seq) : Unfold seq =
-        unfolds |> Seq.mapi (fun offset x -> { i = baseIndex + int64 offset; c = x.EventType; d = x.Data; m = x.Meta; t = DateTimeOffset.UtcNow } : Unfold)
+
+    let mkUnfold compress baseIndex (unfolds: IEventData<_> seq) : Unfold seq =
+        unfolds
+        |> Seq.mapi (fun offset x ->
+            {
+                i = baseIndex + int64 offset
+                c = x.EventType
+                d = x.Data |> if compress then JsonCompressedBase64Converter.Compress else id
```
bartelink: can assign function to temp on L549 though I guess splitting it out is not outright bad!

tests/Equinox.Cosmos.Integration/JsonConverterTests.fs
```diff
@@ -18,20 +19,29 @@ let defaultOptions = FsCodec.SystemTextJson.Options.Create()
...
+            | A x | B x when obj.ReferenceEquals(null, x) -> true
+            | A { embed = x } | B { embed = x } -> obj.ReferenceEquals(null, x)
+        if hasNulls then () else
+
+        let encoded = eventCodec.Encode(None,value)
         let e : Store.Unfold =
             {   i = 42L
                 c = encoded.EventType
-                d = encoded.Data
+                d = JsonCompressedBase64Converter.Compress encoded.Data
```
bartelink: cant this conditionality be driven by a `(compress : bool)` arg on the Property ?

src/Equinox.Cosmos/Cosmos.fs
ylibrach: Unfortunately, at the current time, that's the best (and only) way to do it. I think that in 5.0, they're planning to make JsonElements easier to create. When that happens (I think ETA is Nov), we can switch this logic around a bit.

src/Equinox.Cosmos/Cosmos.fs
ylibrach: Yea I was thinking of doing it but then decided to avoid the extra allocation. But I can definitely switch it to that.

bartelink: are you sure this all collapses in the IL to no allocations though ? (I'd have inlined the pipe and removed the DRY to avoid the question, but really the chances are I'd have made a temp in order to continue to shoehord it onto one line ;) )
It's all good, just remarking

src/Equinox.Cosmos/Cosmos.fs
bartelink: ok - I guess we'll worry about the string concat allocations at that point!

tests/Equinox.Cosmos.Integration/JsonConverterTests.fs
ylibrach: Yep, good point. Changed.

src/Equinox.Cosmos/Cosmos.fs
ylibrach: No worries, I updated it to use a temp function

PR #197, opened 2020-02-19
feat!(CosmosStore): Azure.Cosmos 'V4' support / enable mocking Cosmos APIs

bartelink:
Takes `master` (which uses Cosmos SDK v3 (`Microsoft.Azure.Cosmos`) + makes it use V4 (`Azure.Cosmos`)

The diffs are fairly large between current `master` and the base of this PR; main reason being that Equinox.CosmosStore has cherry-picked the usage of System.Text.Json, leaving only two interesting facilities:
1. using the stillborn `Azure.Cosmos` package
2. introducing some mocking support, which I don't believe anyone has a use for

- [x] port logic
- [x] port to `System.Text.Json`
- [x] Merge #202 (which pushes some code in here out to https://github.com/jet/FsCodec/pull/38)
- [x] Merge #203 (@ylibrach's Client Abstractions work)
- [x] Merge #206 (@ylibrach's Unfolds Compression flag)
- [x] #208 Address `failwith "TODO"` re `eqx dump` functionality 
- [x] #208 Review Internal things like Gateway and other such policy tweaks with a view to aggressively removing and/or de-emphasizing things in V1/V2 that are not being actively used
- [x] #216 Rename to `Equinox.CosmosStore` (with `CosmosStoreConnection`/`Context`/`Category`)
- [x] https://github.com/jet/FsCodec/pull/48 provide path to map between `FsCodec.SystemTextJson` and `FsCodec.NewtonsoftJson`
- [x] release `FsCodec.SystemTextJson` beta so we don't end up having a beta release depending on an `alpha` of something
- [x] Remove `CancellationToken` parameters
- [ ] @ylibrach renaming Equinox.CosmosStore.Core.ContainerGateway to `CosmosContainerAdapter` and any other desired changes
- [ ] @bartelink to release next alpha and validate on a branch in `dotnet-templates`
- [ ] ~Investigate RU consumption discrepancies first observed in #144 - atm there can be a 10% RU consumption increase, which seems (NOTE: wild speculation) to correlate with a lack of anticipated perf improvements when dealing with large spans of small events~ [`master` now as correct RU consumption using V3] now has correct RU consumption(https://github.com/Azure/azure-cosmos-dotnet-v3/issues/616#issuecomment-629813295), so there's reason to believe that this will eventually be righted in the V4 SDK too
[...]

PR #207, opened 2020-03-19
Cosmos SDK 4 code review / cleanup

bartelink:
re #197 #203 Random changes as I review the code

### DIFF
--- a/src/Equinox.Cosmos/Cosmos.fs
+++ b/src/Equinox.Cosmos/Cosmos.fs
@@ -7,9 +7,7 @@ open FsCodec
 open FSharp.Control
 open Serilog
 open System
-open System.IO
 open System.Text.Json
-open System.Text.Json.Serialization
 open System.Threading
 
 /// A single Domain Event from the array held in a Batch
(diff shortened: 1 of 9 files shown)


src/Equinox.Cosmos/Cosmos.fs
```diff
@@ -1179,7 +1165,7 @@ type Discovery =
             UriAndKey (Uri uri, key)
         | _ -> invalidArg "connectionString" "unrecognized connection string format; must be `AccountEndpoint=https://...;AccountKey=...=;`"
 
-type EquinoxCosmosOperationsFactory
```
ylibrach: You really hate the `Equinox` in front, huh? ;)

bartelink: Maybe ;) as mentioned, there;'s nothing with `Equinox` in the type name anywhere else so the removal is driven by an ill-defined consistency instinct....

bartelink: it gets worse :( 

I've pended renaming it to `StoreOperationsFactory.Create() : Equinox.Cosmos.Store.Operations`

Am going to spend some quality time with my thesaurus (and PoeSA book) as any glee you might imagine me having is not here...

bartelink: I'm coming back to thinking it should be called [something including] `Gateway` based on https://martinfowler.com/eaaCatalog/gateway.html

`Equinox.Cosmos.Store` currently has short names and its not a mainstream direct usage thing

if you don't hate the name, then I'd try to find a better name for the one that's in `.Internal` atm ?

bartelink:
Merging this - I'm currently doing a follow-on cleanup exercise, but want to release the alpha from `cosmossdk4` so @ylibrach can test the integration

PR #209, opened 2020-03-24
SQLStreamStore.MsSql: Fix for Initial Append operation failing

Kimserey:
This PR fixes an issue with the current SQLStreamStore implementation preventing new streams to be created.

It also contains fixes for issues with the `samples/Web` project so that it can be tested using it.

---

Fixing the `docker-compose.yml` and the `MySql.Integration` is more involved as it requires to setup a new database (not just spinning up a new container). I'll try to put it together in a different PR.


samples/Web/Startup.fs
```diff
@@ -40,7 +40,9 @@ type App = class end
 type Startup() =
     // This method gets called by the runtime. Use this method to add services to the container.
     static member ConfigureServices(services: IServiceCollection, args: ParseResults<Arguments>) : unit =
-        services.AddMvc().SetCompatibilityVersion(CompatibilityVersion.Latest) |> ignore
+        services
+            .AddMvc(fun opt -> opt.EnableEndpointRouting <- false)
```
Kimserey: `opt.EnableEndpointRouting  <- false` needs to be specified otherwise the sample hit the following runtime error:

```
Application startup exception: System.InvalidOperationException: Endpoint Routing does not support 'IApplicationBuilder.UseMvc(...)'. To use 'IApplicationBuilder.UseMvc' set 'MvcOptions.EnableEndpointRouting = false' inside 'ConfigureServices(...).
   at Microsoft.AspNetCore.Builder.MvcApplicationBuilderExtensions.UseMvc(IApplicationBuilder app, Action`1 configureRoutes)
   at Microsoft.AspNetCore.Builder.MvcApplicationBuilderExtensions.UseMvc(IApplicationBuilder app)
   at Web.Startup.Configure(IApplicationBuilder app, IHostEnvironment env) in /Users/k0l01ap/repos/equinox-kim/equinox/samples/Web/Startup.fs:line 92
   at Web.Program.createWebHostBuilder@15-1.Invoke(IApplicationBuilder app) in /Users/k0l01ap/repos/equinox-kim/equinox/samples/Web/Program.fs:line 15
   at Microsoft.AspNetCore.Hosting.WebHostBuilderExtensions.<>c__DisplayClass0_0.<Configure>b__0(WebHostBuilderContext _, IApplicationBuilder app)
   at Microsoft.AspNetCore.Hosting.WebHostBuilderExtensions.<>c__DisplayClass2_1.<Configure>b__2(IApplicationBuilder app)
   at Microsoft.AspNetCore.Hosting.DelegateStartup.Configure(IApplicationBuilder app)
   at Microsoft.AspNetCore.Mvc.Filters.MiddlewareFilterBuilderStartupFilter.<>c__DisplayClass0_0.<Configure>g__MiddlewareFilterBuilder|0(IApplicationBuilder builder)
   at Microsoft.AspNetCore.HostFilteringStartupFilter.<>c__DisplayClass0_0.<Configure>b__0(IApplicationBuilder app)
   at Microsoft.AspNetCore.Hosting.WebHost.BuildApplication()
```

src/Equinox.SqlStreamStore/SqlStreamStore.fs
```diff
@@ -134,7 +134,7 @@ module private Write =
     let private writeEventsAsync (log : ILogger) (conn : IEventStoreConnection) (streamName : string) (version : int64) (events : EventData[])
         : Async<EsSyncResult> = async {
         try
-            let! wr = conn.AppendToStream(StreamId streamName, int version, events) |> Async.AwaitTaskCorrect
+            let! wr = conn.AppendToStream(StreamId streamName, (if version = -1L then ExpectedVersion.NoStream else int version), events) |> Async.AwaitTaskCorrect
```
Kimserey: Actual fix for the issue.

samples/Web/Controllers/TodosController.fs
```diff
@@ -6,6 +6,7 @@ open TodoBackend
 
 type FromClientIdHeaderAttribute() = inherit FromHeaderAttribute(Name="COMPLETELY_INSECURE_CLIENT_ID")
 
+[<CLIMutable>]
```
Kimserey: `CLIMutable` attribute is necessary else we hit the following exception:

```
System.NotSupportedException: Deserialization of reference types without parameterless constructor is not supported. Type 'TodoBackend.Controllers.TodoView'
```

samples/Web/Startup.fs
```diff
@@ -86,6 +88,5 @@ type Startup() =
         if env.IsDevelopment() then app.UseDeveloperExceptionPage() |> ignore
         else app.UseHsts() |> ignore
 
-        app.UseHttpsRedirection()
-            .UseCors(fun x -> x.WithOrigins([|"https://www.todobackend.com"|]).AllowAnyHeader().AllowAnyMethod() |> ignore)
-            .UseMvc() |> ignore
\ No newline at end of file
+        app.UseCors(fun x -> x.AllowAnyOrigin().AllowAnyHeader().AllowAnyMethod() |> ignore)
```
Kimserey: Right now I can't get the `todobackend.com` website to work with `WithOrigins([|"https://www.todobackend.com"|])` so since this is not the _main focus_ of the sample I'd suggest we make it more flexible by allowing any origin.

samples/Web/Startup.fs
bartelink: Wondering if adding `UseRouting` like https://github.com/jet/dotnet-templates/blob/master/equinox-web-csharp/Web/Startup.cs#L22 can do work instead? (I ran into it and did it that way - I have no fixed opinion on which fix is better but ideally they'd be in sync)

src/Equinox.SqlStreamStore/SqlStreamStore.fs
bartelink: Cool - trusting it doesnt break MySql and Postgres ;)

samples/Web/Controllers/TodosController.fs
bartelink: I believe this is best remediated via `AddNewtonsoftJson` ? https://github.com/jet/dotnet-templates/blob/master/equinox-web-csharp/Web/Startup.cs#L39

samples/Web/Startup.fs
bartelink: relaxing is not the end of the world but per my quick tests it seems to work in https://github.com/jet/dotnet-templates/blob/master/equinox-web-csharp/Web/Startup.cs#L27 ?

samples/Web/Controllers/TodosController.fs
Kimserey: Yeah that seems to work, I had to add `Microsoft.AspNetCore.Mvc.NewtonsoftJson`.

samples/Web/Startup.fs
Kimserey: I am not sure I can't get it to work, I hit CORS issues if I try to use the non HTTPS endpoint.
And I hit ERR certificate if I want to use the HTTPS endpoint.

In order to get the sample to work,  I had to remove the HTTPS redirection and allow all origins.

samples/Web/Startup.fs
Kimserey: Yeah that works, thanks! I'll make the changes.

Kimserey:
@bartelink I've made the changes to align the sample with the dotnet template for newtonsoft json and for the aspnet core 3.x routing.

I still couldn't get CORS to work.

Let me know what you think. thanks!

samples/Web/Startup.fs
```diff
@@ -40,7 +40,10 @@ type App = class end
 type Startup() =
     // This method gets called by the runtime. Use this method to add services to the container.
     static member ConfigureServices(services: IServiceCollection, args: ParseResults<Arguments>) : unit =
-        services.AddMvc().SetCompatibilityVersion(CompatibilityVersion.Latest) |> ignore
+        services
+            .AddMvc()
+            .AddNewtonsoftJson()
```
bartelink:
```suggestion
            .AddNewtonsoftJson() // TODO swap to FsCodec.SystemTextJson when it's ready https://github.com/jet/FsCodec/pull/38
```

samples/Web/Startup.fs
bartelink: cool, thanks

samples/Web/Controllers/TodosController.fs
bartelink: cool, thanks

samples/Web/Startup.fs
bartelink: 1) if the code needs to change, we'll chang it, but the correct/intended code absolutely should remain in a comment or an ifdef
2) dotnet dev-certs can fix HTTPS, but a comment would be friendlier...

Kimserey: After offline discussion, the HTTPS problem can be fixed with dotnet dev-certs:

```
❯ dotnet dev-certs https --clean
❯ dotnet dev-certs https --trust
```

Thanks!

PR #214, opened 2020-04-03
small wording changes

swrhim:
Something I'm not sure what the wording should be is stores -> reactors : trigger. I'm not sure how to make this explicit. Thoughts?


diagrams/context.puml
```diff
@@ -73,12 +73,12 @@ services   <-R-> stores     : Query/Transact\n[in terms of events]
 stores       ..> publishers : are followed by
 stores       ..> reactors   : trigger
 
-reactors     --> services   : feed back into
+reactors     --> services   : feed data back into
```
bartelink, line 76: I'm intentionally being vague/general here - a reaction is driving a decision and/or could be triggering idempotent command processing driven by a process manager. I'm not sure what a better term for that might be, but _data_ seems overly specific and makes it sound more like an Ingester (which probably can benefit from being cast as doing something straightforward)

diagrams/context.puml
```diff
@@ -73,12 +73,12 @@ services   <-R-> stores     : Query/Transact\n[in terms of events]
...
 reactors     ..> outputs    : generate complex\nnotifications for
 reactors    .U.> caches     : maintain views in
 
 publishers   ..> outputs    : emit complex notifications to
-publishers   ..> outputs    : render some events to
```
bartelink, line 81: here, I'm trying to convey that it's being selective in what it opts to render (whereas for complex notifications, it's publishing every one that's been prepared by the "Compose Notifications" umbrella thing (which is another reason why one might not be feeding just `data` above))

bartelink:
>Something I'm not sure what the wording should be is stores -> reactors : trigger. I'm not sure how to make this explicit. Thoughts?

I updated the label to be *fresh writes trigger* since...

Can you review and/or rebase this based on recent updates?

diagrams/context.puml
swrhim: Right now, I think it's too vague. Saying even feed decision back into the service would make this a lot clearer

diagrams/context.puml
swrhim: how about, render filtered events to?

diagrams/context.puml
bartelink: https://github.com/jet/equinox/blob/diag/DOCUMENTATION.md now says "manage follow-on reactions via"

diagrams/context.puml
bartelink: image

swrhim:
@bartelink If we can resolve the last comment, I'd want to abandon this PR and have you update the change directly on master. Lots of conflicts and doesn't seem worth it to resolve them for small changes

PR #217, opened 2020-04-27
Improve readability of unrendered DOCUMENTATION.md

xandkar:
Some crazies, like myself, prefer to read files in their text editor
rather than in a browser, in which case it helps when lines aren't
1,000,000+ characters long :)

There are a few drive-by for-rendering fixes as well, namely changing:

a) foo
b) bar
...

and

i) foo
ii) bar
...

which are rendered as continuous text (rather than list) by GitHub into

1. foo
2. bar
...

which is rendered as an ordered list.

I also aligned a couple of small tables, but left the others alone for
now (too much work).

Some (example) code lines were also split.

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -1,36 +1,75 @@
 # Documentation
 
-Please refer to the [FAQ](README.md#FAQ), [README.md](README.md), the [Ideas](#ideas) and the [Issues](https://github.com/jet/equinox/issues) for background info on what's outstanding (aside from there being lots of room for more and better docs).
+Please refer to the [FAQ](README.md#FAQ), [README.md](README.md), the
+[Ideas](#ideas) and the [Issues](https://github.com/jet/equinox/issues) for
+background info on what's outstanding (aside from there being lots of room for
+more and better docs).
 
 # Background reading
 
-While there is no canonical book on doing event-sourced Domain Modelling, there are some must read books that stand alone as timeless books in general but are also specifically relevant to the considerations involved in building such systems:
+While there is no canonical book on doing event-sourced Domain Modelling, there
+are some must read books that stand alone as timeless books in general but are
+also specifically relevant to the considerations involved in building such
+systems:
 
-_Domain Driven Design, Eric Evans, 2003_ aka 'The Blue Book'; not a leisurely read but timeless and continues to be authoritative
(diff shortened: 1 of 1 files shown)


DOCUMENTATION.md
```diff
@@ -447,54 +614,113 @@ type Service internal (resolve : ClientId -> Equinox.Stream<Events.Event, Fold.S
...
+abstractions; the 2 files can tell many of the thousands of words about to
+follow!
 
 #### Stream Members
 
 ```fsharp
 type Equinox.Stream(stream : IStream<'event, 'state>, log, maxAttempts) =
 StoreIntegration
-    // Run interpret function with present state, retrying with Optimistic Concurrency
+    // Run interpret function with present state, retrying with Optimistic
```
bartelink, line 718: the fact this is one line is very dear to me...

DOCUMENTATION.md
```diff
@@ -447,54 +614,113 @@ type Service internal (resolve : ClientId -> Equinox.Stream<Events.Event, Fold.S
...
+    // Concurrency
     member __.Transact(interpret : State -> Event list) : Async<unit>
 
-    // Run decide function with present state, retrying with Optimistic Concurrency, yielding Result on exit
+    // Run decide function with present state, retrying with Optimistic
```
bartelink, line 722: again, I trued to make this a pithy one line, which this destroys the spirit of

DOCUMENTATION.md
```diff
@@ -803,9 +1210,11 @@ let decide (context, command) state : int * Events.Event list =
 
 type Service internal (resolve : ClientId -> Equinox.Stream<Events.Event, Fold.State>) =
 
-    // Given the supplied context, attempt to apply the command for the specified clientId
+    // Given the supplied context, attempt to apply the command for the
```
bartelink, line 1213: usual concern...

DOCUMENTATION.md
xandkar: Will fix

DOCUMENTATION.md
xandkar: Can `\n` go after `,` or do you want the whole thing back to original?

xandkar: that is, `Concurrency, yielding ` --> `Concurrency,\nyielding`

bartelink: open to splitting on comma

PR #222, opened 2020-06-12
feat(EventStoreDb): Add AccessStrategy.CustomOrigin

bartelink:
Provides a `Equinox.EventStore.AccessStrategy.CustomOrigin` which would be useful in the context of https://github.com/jet/dotnet-templates/pull/40 to enable a custom lookback on a long-running stream (without employing the snapshots strategy).

🤔 Still mulling over whether this is truly a good idea to expose in Equinox


bartelink:
See also #263 - similar considerations make me not want to expose it until someone has a use case that we can't provide a clean workaround for

PR #227, opened 2020-07-04
SSS Update to 1.2.0

Rajivhost:
update `SqlStreamStore` libs (except PostGreSQL) to `1.2.0`


bartelink:
Thanks!
I'll likely hold off on merging this until the postgres equivalent is available too though unless someone is actually blocked....

bartelink:
Thanks for the contribution! This will be in the 4.x release line.
(Per DDD-CQRS-ES, the SqlStreamStore.Postgres lib will remain tagged a beta to reflect the fact that it's not a complete validated release)

PR #228, opened 2020-07-09
Add AsyncBatchingGate

bartelink:
Helper primitive that can be used to constrain contention where multiple threads are targeting a single upstream.

The specific need at hand is where multiple threads in a projector are transacting against the same aggregate. In such a case the duelling transactions trigger `Equinox.Cosmos` `Resync` operations / ESDB `WrongExpectedVersion` which increases load for the upstream, in addition to increasing latency due to having to retry.

-----
Like `AsyncCacheCell` (although that's slightly more justifiable given its used in the impl of Equinox.Cosmos), these helpers are generic; the happen to be useful in typical Concurrent Transaction processing work, but are not "part of Equinox". Hosting this in Equinox.Core is thus not a long term plan - ideally these primitives ultimately end up with their own home at some point (not dissimilar to how Equinox.Codec moved out to FsCodec).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ The `Unreleased` section name is replaced by the expected version of next releas
 
 ### Added
 
+- Add `AsyncBatchingGate` [#228](https://github.com/jet/equinox/pull/228)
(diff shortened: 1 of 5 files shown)


bartelink:
(I'm intending to merge this and include in a 2.2.0 release when that happens - atm it's being validated internally)

PR #229, opened 2020-07-09
Simplify AsyncCacheCell

bartelink:
The initial impl of AsyncCacheCell passes the tests and fulfilled the need in the context that it was written (an initialization gating device in `Equinox.Cosmos`)

However the `PeekIsValid` perf tweak is a messy ad-hoc thing, so as precursor to https://github.com/jet/equinox/pull/228 I clarified and simplified the logic


bartelink:
Atm, a this does not achieve any particular behavior advantage, I have no plans to put this onto the 2.x release line and instead will leave it on master to be validated as part of the 3.x release series

PR #234, opened 2020-08-28
Removed BlockingCollection with a lock-free queue for AsyncBatch

thednaz:


bartelink:
Closing as discussed out of band (the PR offers a perf benefit, but the cognitive load and LOC increment does not justify it)

PR #245, opened 2020-09-29
Update `eqx init` to support Serverless re #244

### LINKED ISSUE #244: Cosmos: Add Serverless account type support
`eqx init cosmos` currently does not support the provisioning of the equinox container on a Serverless account type. By enabling support for serverless account types this will allow lower cost development and testing on Azure based Cosmos DB Accounts.

### Serverless Requirements:
- When provisioning a serverless containers you are not allowed to configure the RUs for the container
- When using a serverless account you cannot read the provisioned RUs for a container or database and doing so returns an error.
- You cannot create a shared throughput database in a serverless account and doing so returns an error

### Relevant Documents:
- https://docs.microsoft.com/en-us/azure/cosmos-db/serverless
- https://docs.microsoft.com/en-us/azure/cosmos-db/throughput-serverless

OmnipotentOwl:
- [x] #244 re being able to provision on serverless account type.
- [x] Tidy

resolves #244

### DIFF
--- a/tools/Equinox.Tool/Program.fs
+++ b/tools/Equinox.Tool/Program.fs
@@ -13,6 +13,8 @@ open System
 open System.Net.Http
 open System.Threading
 
+type Provisioning = Equinox.CosmosStore.Core.Initialization.Provisioning
+
 let [<Literal>] appName = "equinox-tool"
 
 [<NoEquality; NoComparison>]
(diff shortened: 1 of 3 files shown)


bartelink:
Interesting to look at how its done in V4 - maybe if no RUs are specified, it should not touch the RUs?
https://github.com/jet/equinox/blob/cosmossdk4/src/Equinox.CosmosStore/CosmosStore.fs#L354

In general this looks great - I want to do some other merges/syncs with v2 branches before I merge this but it should be later today :fingerscrossed:

tools/Equinox.Tool/Program.fs
```diff
@@ -314,11 +325,19 @@ module CosmosInit =
...
         match iargs.TryGetSubCommand() with
         | Some (InitArguments.Cosmos sargs) ->
-            let rus, skipStoredProc = iargs.GetResult(InitArguments.Rus), iargs.Contains InitArguments.SkipStoredProc
-            let mode = if iargs.Contains InitArguments.Shared then Provisioning.Database rus else Provisioning.Container rus
-            let modeStr, rus = match mode with Provisioning.Container rus -> "Container",rus | Provisioning.Database rus -> "Database",rus
+            let skipStoredProc = iargs.Contains InitArguments.SkipStoredProc
             let client,dName,cName = conn log sargs
-            log.Information("Provisioning `Equinox.CosmosStore` Store at {mode:l} level for {rus:n0} RU/s", modeStr, rus)
+            let mode = (CosmosInitInfo iargs).ProvisioningMode
+            match mode with
```
bartelink, line 331: Arguably, this logging code could move up into `CosmosInitInfo` - splitting parsing/logging from the doing, but in general in the tool, I don't follow [the rules the templates do](https://github.com/jet/dotnet-templates#module-args) to the same degree

OmnipotentOwl:
Confirmed the functionality of serverless provisioning for after the merges. One thought I did have around this when I was implementing the provisioning logic was that the aux container setup may also need the same updates to support serverless. I also didn't see right away if the aux container provisioning supported the shared Db RUs or not.

bartelink:
Thanks for all the work and for validating; I'll merge it and cherry-pick it onto the `v2` branch for the next release (`2.2.`x)

The aux provisioning is driven by propulsion.tool in the propulsion repo.

In general, if you're running a CFP somewhere with state in an aux container, you can be pretty sure there's continual usage (not sure how triggers run with Azure Functions driven off CFPs though)

As a result, I'd say you want to stick with non-serverless allocation for aux containers, in general.

The V3 API has a variety of scenarios for how one can consume from a Change Feed (i.e. its not hard wired to storing state in an aux container in the same way).

bartelink:
(☝️ Second is the relevant commit on `v2` that cherry-picks this change for inclusion in `2.2.x`)

PR #248, opened 2020-10-02
CosmosStore: Enable uncompressed unfolds as in v2 #249

bartelink:
Port of @ylibrach's https://github.com/jet/equinox/pull/206 to work with `Newtonsoft.Json`
Enables a `CosmosCategory` to inhibit compressing of Unfolds (Snapshot bodies) stored in Tip by adding `, compressUnfolds=false`

CAUTION: Use with care: readers with older versions of Equinox will choke on uncompressed values


bartelink:
First official release of this is on `v2` branch - this is effectively a cherrypick of that

Issue #232, opened 2020-08-10
V3 checklist


bartelink:
Breaking changes and other considerations for V2 vs V3:

- Equinox.EventStore V2 will depend on EventStore.ClientApi V5.x whereas V3 will depend on 20.x
- Equinox.Cosmos V2 will depend on Microsoft.Azure.DocumentDb.Core V2.x whereas V3 will depend on Azure.Cosmos (aka V4)
- V2 supports full FW and/or netcore 1.x, with tools on 2.x - V3 may opt to remove full FW support and/or have tools depend on 5.x

Breaking change ideas:
- consider moving `maxAttempts` and associated throwing behavior from `Stream` to `Transact` (1. for memory testing you want a higher count 2. to force consideration at the right point in the learning experience)

PR #250, opened 2020-10-08
Fixed a broken URL

dharmaturtle:


dharmaturtle:
Also, the link to Equinox.CosmosStore's nuget package is broken. Should I point it to Equinox.Cosmos for now until te nuget package gets updated?

image

bartelink:
Thanks for spotting (and taking the time to fix!) this.

Regarding the CosmosStore vs Cosmos naming, this is referenced in https://github.com/jet/equinox/issues/232 and https://github.com/jet/equinox/pull/197

Short answer: Yes, use the `.Cosmos` package (the docs on the v2 branch are in sync with that naming)

----
Longer answer: 

Atm, `master` has a CosmosStore using the `Microsoft.Azure.Cosmos` aka v3 package under the hood, and `v2` has a `.Cosmos` which uses `Microsoft.Azure.DocumentDb.Client.Core` aka v2. There is an alpha of a CosmosStore package on nuget, but thats from the pull request branch above (it uses `Azure.Cosmos` preview nuget aka v4).

Either way, that `.CosmosStore` package has some breaking changes (better naming, _as well as significant efficiency gains_), and the intention is to provide a reasonable upgrade path.

It's likely that, at the point that the `CosmosStore` package enters rc, that I'll re-add the .Cosmos package source (copied from whatever is on the v2 branch) onto the master branch and each release will include both nugets (without breaking changes to the .Cosmos one). This will allow people to use them side-by-side during the course of switching larger codebases over aggregate-by-aggregate.

We'll hopefully get the README in sync when that happens!

bartelink:
Aside: If this is for local experimentation, then replacing:

`<!--    <PackageReference Include="Equinox.Cosmos" Version="2.3.0-rc1" />-->`

with

`<ProjectReference Include="..\..\..\equinox\src\Equinox.CosmosStore\Equinox.CosmosStore.fsproj" />`

can be used to do local experiments

(Alternately if you add a reference to `CosmosStore.fs` into your project you can simply inline it - one of the key motivations for having most projects contain everything in a single file is to allow people to work around bugs by inlining parts locally)

PR #262, opened 2020-11-09
Switch fsx files to nuget references

voronoipotato:
This allows people to try out Equinox by just running the FSX alone. We might want to specify a version number in the #r references here but it works great.


bartelink:
Thanks; good thinking.

I'm thinking this should be done consistently in all the `.fsx`s (potentially with the existing file-based refs in a `#ifdef` to enable people to play what-if when spiking tweaks)

bartelink:
(This has not been forgotten; the blocker is that we've yet to publish V3 nugets that are in sync with the code on `master`)

bartelink:
have published `3.0.0-beta.1`, which is in sync with code on `master` so this can now work...

bartelink:
(There is a 3.0.0 release on Nuget now)

bartelink:
Included in #309

PR #263, opened 2020-11-15
Expose pre-state Version via TransactEx

bartelink:
At present, the `Equinox.Decider` interface only exposes the outgoing `Version`, which means the only way to determine the incoming Version is by indirect means such as:
- checking the `Index` of the freshest event (can be incorrect if an Event that's not in the DU is encountered)
- inferring it based on the post-`Version` passed to the `mapResult` (won't work well with transmute operations or rolling snapshots)

This proposal provides the incoming extended state via a new `TransactEx` API, enabling one to process the value directly.

As this is only for V3, it makes sense to remove `TransactAsyncEx` as both APIs are likely extremely rare, and having 2 mysteriously named functions with complex signatures is rarely better than one.

cc @dharmaturtle

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,8 +44,12 @@ The `Unreleased` section name is replaced by the expected version of next releas
 - Remove separated `Backend` project from examples (support for architecturally separating all domain logic from Equinox and Domain Service logic from Concrete Stores remains)
 - Revise semantics of Cart Sample Command handling
 - Simplify `AsyncCacheCell` [#229](https://github.com/jet/equinox/pull/229)
+- Extend `Decider.TransactEx` signature to expose `ISyncContext` to `decide` functions [#263](https://github.com/jet/equinox/pull/263)
 
 ### Removed
+
+- Remove `Decider.TransactAsyncEx` (`TransactEx` provides superset) [#263](https://github.com/jet/equinox/pull/263)
+
 ### Fixed
 
 - change `createAttemptsExhaustedException` to allow any `exn`-derived `type` [#275](https://github.com/jet/equinox/pull/275)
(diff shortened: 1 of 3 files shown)


src/Equinox/Equinox.fs
```diff
@@ -10,42 +10,50 @@ type MaxResyncsExhaustedException(count) =
 /// Central Application-facing API. Wraps the handling of decision or query flows in a manner that is store agnostic
 type Stream<'event, 'state>
     (   log, stream : IStream<'event, 'state>, maxAttempts : int,
-        [<Optional; DefaultParameterValue(null)>] ?createAttemptsExhaustedException,
+        [<Optional; DefaultParameterValue(null)>] ?createAttemptsExhaustedException : int -> exn,
```
bartelink, line 13: this signature fix should be a separate PR, and probably should only happen in V3
(the restriction for the exception to inherit from `MaxResyncsExhaustedException` was unintentional)

bartelink:
This PR is an example to illustrate an approach in a discussion on the DDD-CQRS-ES Slack - my instinct is not to add this until we validate a scenario for which this truly is the best solution.

bartelink:
@dharmaturtle I'll likely take the cleanup aspects of this PR, and the FAQ from this and merge it after #272 for inclusion in the V3 RC

As noted previously, any concept that gets exposed in the public API has a cost in terms of adding stuff for readers of the code to traverse. That doesn't always make it a bad thing - sometimes naming or mentioning the concept can help people's thinking even if they opt not to actually use the facility in the end.

Please let me know if you've found a use case for needing to know the pre-version or have any other thoughts on this.

dharmaturtle:
I've thrown pre-release binaries into source control before. I don't think it's a bad practice, and [neither does Mark Seemann](https://blog.ploeh.dk/2014/01/29/nuget-package-restore-considered-harmful/). There's no particular rush :)

bartelink:
Thats an interesting idea; in this instance it's more a concern of whether the feature should be added in the first instance than the release vehicle itself though.

If you wanted/needed to trial it, PR builds yield [alpha-badged] artifacts and can be released, though we try not to do that too often in order to avoid stressing nuget's performance.

On re-reading, I'm beginning to like the OP suggestion of exposing it as a breaking change in [`TransactAsyncEx`](https://github.com/jet/equinox/blob/master/src/Equinox/Equinox.fs#L47) - that provides the escape hatch just in case, without suggesting that it's part of any mainline API.

src/Equinox/Equinox.fs
bartelink: opted to leave it here

PR #266, opened 2020-11-23
Add CosmosStore.Prometheus

bartelink:
Adds new `Equinox.CosmosStore.Prometheus` package which publishes metrics to Prometheus via `prometheus-net`.

- Takes the present spartan stats gather facility (which some dotnet-templates use to dump basic traffic stats per-minute), and generalizes the helpers so they are shared with the new package
- applies minor naming/consistency improvements in the Equinox.CosmosStore impl resulting from this work


bartelink:
Will likely be backported to v2 branch, with equivalent changes applied to the other drivers too

PR #270, opened 2020-12-14
Add RollForward for Equinox.Tool

bartelink:
In order to allow usage of tool where only SDK >= 5 is installed in a given context


bartelink:
via https://twitter.com/dsymetweets/status/1333818089809580035 via F# weekly ;)

PR #272, opened 2021-01-26
Rename Equinox.Stream -> Decider

bartelink:
Finally, a nice name for it; based on thinking shared by @thinkbeforecoding on DDD-CQRS-ES slack

### DIFF
--- a/DOCUMENTATION.md
+++ b/DOCUMENTATION.md
@@ -622,7 +622,7 @@ let create resolve : Service =
 The most terse walkthrough of what's involved in using Equinox to do a
 Synchronous Query and/or Execute a Decision process is in the
 [Programming Model section](#programming-model). In this section we’ll walk
-through how one implements common usage patterns using the Equinox `Stream` API
+through how one implements common usage patterns using the Equinox `Decider` API
 in more detail.
 
 ## ES, CQRS, Event-driven architectures etc
@@ -686,7 +686,7 @@ At a high level we have:
 
 ## Programming Model walkthrough
 
-### Flows and Streams
+### Flows and Deciders
 
 Equinox’s Command Handling consists of < 200 lines including interfaces and
 comments in https://github.com/jet/equinox/tree/master/src/Equinox - the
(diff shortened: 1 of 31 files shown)


thinkbeforecoding:
To give more information about the reason of the name:
- Stream is linked the the specificity of the storage in an Event Sourcing system
- Aggregate (in DDD terminology) is too much about the shape (which tend to disappear in an Event Sourcing system)

Finally the the role of this is to take decision (in autonomie, not relying on other information).
it signature is:
decide : Command -> State -> Event[]
evolve: State -> Event -> State
initialState: State

it can be made a pure decider by having State = Event[]:
initialState: []
evolve: State -> Event -> Event[]  // just append event to list

and the decider becomes (but avoid it, it's less efficient at runtime, but it's equivalent conceptually):
decide: Command -> history:Event[] -> Event[]

Here is the reason to call it Decider

bartelink:
I'm intending to merge this prior to the first RC release of V3

DOCUMENTATION.md
```diff
@@ -696,9 +696,9 @@ elements you'll touch in a normal application are:
   internal implementation of Optimistic Concurrency Control / retry loop used
   by `Stream`. It's recommended to at least scan this file as it defines the
```
thinkbeforecoding, line 697: You left one here 😉

DOCUMENTATION.md
```diff
@@ -696,9 +696,9 @@ elements you'll touch in a normal application are:
...
   Transaction semantics everything is coming together in service of.
-- [`type Stream`](https://github.com/jet/equinox/blob/master/src/Equinox/Equinox.fs#L11) -
+- [`type Decider`](https://github.com/jet/equinox/blob/master/src/Equinox/Decider.fs#L11) -
   surface API one uses to `Transact` or `Query` against a specific stream
-- [`type Target` Discriminated Union](https://github.com/jet/equinox/blob/master/src/Equinox/Equinox.fs#L42) -
+- [`type Target` Discriminated Union](https://github.com/jet/equinox/blob/master/src/Equinox/Decider.fs#L42) -
   used to identify the Stream pertaining to the relevant Aggregate that
   `resolve` will use to hydrate a `Stream`
```
thinkbeforecoding, line 703: Same

thinkbeforecoding:
Seems to small comments changes, but good!
The Decider term start to be used in the eventsourcing community (at least on the DDD + EventModeling slacks)

DOCUMENTATION.md
bartelink: that `Target` was actually way out of date, thanks for flagging 😊

DOCUMENTATION.md
bartelink: Thanks for reviewing and catching; That was actually only the head of a seam, lots of fixes pushed 😊

PR #278, opened 2021-02-26
Cosmos: Use 3.13; Dispose FeedResponse fixes #225

### LINKED ISSUE #225: Cosmos: Dispose queries/results if releasing against Cosmos SDK V3 instead of V4
NOTE This does not apply to the present 2.x release series

The `Microsoft.Azure.Cosmos` APIset [introduces `IDisposable` on queries and responses](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/1638) - the current `master`/`main` branch uses this 'V3' API and hence should be updated to ensure memory usage is correct and/or optimal.

I believe the `Azure.Cosmos` (aka V4) SDK makes this moot by exposing the features as `IAsyncEnumerable`. Thus, assuming we don't release an impl targeting V3 and instead leapfrog from SDK v2 to V4 via https://github.com/jet/equinox/pull/197 we can avoid needing any code changes.

bartelink:
Fixes #225

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,12 @@ The `Unreleased` section name is replaced by the expected version of next releas
 ### Changed
 
 - `CosmosStore`: Default Connection Mode now `Direct` to align with V3 SDK (previous default was `Gateway` to match V2 SDK) [#281](https://github.com/jet/equinox/pull/281)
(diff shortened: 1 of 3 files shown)


bartelink:
Should probably update to 3.17 before merging (esp if that hits GA; its currently in `-preview`)

PR #283, opened 2021-03-22
Equinox.Tool: Fix incorrect Azure.Cosmos.Direct ref

bartelink:
Fix for 3.0.0-beta.3


bartelink:
A workaround if using beta build of `Equinox.Tool` is to run with `-m gateway`

PR #280, opened 2021-03-03
Cosmos: Merge Connector+Connection as CosmosStoreClient

bartelink:
Merges two concepts to better align with `Azure.Core` design principles which will align with `CosmosClient` V4 design.
Fixes #284

PR #291, opened 2021-06-11
Fix accidental reversion of #281

bartelink:
Reverts overwrite of changes in #281 in #280

Typically, this will not surface as a bug in normal usage as `dotnet-templates` explicitly specifies the `mode`


bartelink:
@deviousasti I'm going to merge this, post a follow-up PR, and then stage a 3.0.1 release with a minor breaking change that simplifies the templates.

PR #296, opened 2021-11-02
MemoryStore: Fix Version for post-context in Equinox.Decider.TransactEx

bartelink:
Corrects `MemoryStore`'s computation of the post-`Version` for use by `Equinox.Decider.TransactEx`
Refactors conflict case path to remove ugly use of `Exception` that I once assumed to be unavoidable

### DIFF
--- a/src/Equinox.MemoryStore/MemoryStore.fs
+++ b/src/Equinox.MemoryStore/MemoryStore.fs
@@ -12,11 +12,11 @@ exception private WrongVersionException of value : obj
 
 /// Internal result used to reflect the outcome of syncing with the entry in the inner ConcurrentDictionary
 [<NoEquality; NoComparison>]
-type ConcurrentDictionarySyncResult<'t> = Written of 't | Conflict
+type ConcurrentDictionarySyncResult<'Format> = Written of FsCodec.ITimelineEvent<'Format>[] | Conflict
 
 /// Response type for VolatileStore.TrySync to communicate the outcome and updated state of a stream
 [<NoEquality; NoComparison>]
-type ConcurrentArraySyncResult<'t> = Written of 't | Conflict of 't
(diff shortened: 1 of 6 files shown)


src/Equinox.MemoryStore/MemoryStore.fs
```diff
@@ -67,47 +67,41 @@ module private Token =
             version = int64 eventCount }
     let (|Unpack|) (token: StreamToken) : Token = unbox<Token> token.value
     /// Represent a stream known to be empty
-    let ofEmpty streamName initial = streamTokenOfEventCount streamName 0, initial
-    let tokenOfArray streamName (value: 'event array) = Array.length value |> streamTokenOfEventCount streamName
-    let tokenOfSeq streamName (value: 'event seq) = Seq.length value |> streamTokenOfEventCount streamName
-    /// Represent a known array of events (without a known folded State)
-    let ofEventArray streamName fold initial (events: 'event array) = tokenOfArray streamName events, fold initial (Seq.ofArray events)
-    /// Represent a known array of Events together with the associated state
-    let ofEventArrayAndKnownState streamName fold (state: 'state) (events: 'event seq) = tokenOfSeq streamName events, fold state events
```
bartelink, line 76: note this computes an incorrect token as the event count is derived from the new events, not the post-state value of the events for this stream

bartelink:
Spotted during code review as I wondered whether SELECT was broken - it wasn't...

Aside: cctual PEBKAC that made me find it
```
Estimated e -> {state with estimates = [| for x in state.estimates do if x.kind <> e.kind then x; e|] } 
```
^ does not do what I mean

src/Equinox.MemoryStore/MemoryStore.fs
bartelink: also an example of how confusing egregious use of `|>` is never a good thing (I try very hard not to do that as I'm conscious of it, but I evidently thought I was being very clever)

PR #301, opened 2021-11-12
@oskardudycz @thinkbeforecoding fixes re #299

bartelink:
Was a bit hasty in merging - this applies some excellent suggestions from @oskardudycz


README.md
```diff
@@ -832,8 +832,10 @@ The single best treatment of the concept of a Decider that's online at present i
 
 #### In Equinox
 
-The Equinox `type Decider`, which exposes an [API that covers the needs of making Consistent Decisions against a State derived from Events on a Stream](
-https://github.com/jet/equinox/blob/master/src/Equinox/Decider.fs#L22-L56)
+The Equinox `type Decider`, exposes an [API that covers the needs of making Consistent Decisions against a State derived from Events on a Stream](
```
ragiano215, line 835:
```suggestion
The Equinox `type Decider` exposes an [API that covers the needs of making Consistent Decisions against a State derived from Events on a Stream](
```

README.md
```diff
@@ -832,8 +832,10 @@ The single best treatment of the concept of a Decider that's online at present i
...
+https://github.com/jet/equinox/blob/master/src/Equinox/Decider.fs#L22-L56). At a high level, we have:
+- `Transact*` functions - these run a decision function that may result in a change to the State, including management of the retry cycle when a consistency violation occurs during the syncing of the state with the backing store (See [Optmimistic Concurrency Control](https://en.wikipedia.org/wiki/Optimistic_concurrency_control)). Some variants can also yield an outcome to the caller after the syncing to the store has taken place.
+- `Query*` functions - these run a render function propjecting from the State the Decider manages (but can't mutate it or trigger changes). The concept of [CQRS](https://martinfowler.com/bliki/CQRS.html) is a consideration here - using the Decider to read state should not be a default approach (but equally should not be considered off limits).
```
ragiano215, line 838:
```suggestion
- `Query*` functions - these run a render function projecting from the State that the Decider manages (but can't mutate it or trigger changes). The concept of [CQRS](https://martinfowler.com/bliki/CQRS.html) is a consideration here - using the Decider to read state should not be a default approach (but equally should not be considered off limits).
```

README.md
```diff
@@ -854,12 +856,12 @@ _When applying the concept of a Decider to event sourcing, the consistency requi
 #### The ingredients
 
 With Deciders in general, and Equinox in particular, the following elements are involved:
-- a `State` type, on which we base decisions, which can be updated by the consequences of a decision
-- a `decide` function, which is presented a `State`, and returns a decision, which we use to update the State, if, and only if there is no concurrency conflict when applying them
-- an `initial` State, from which we start
-- an `Event` type (think a Discriminated union)
-- the Events are maintained as an ordered list of events (could be an in memory array, a stream in ESDB, documents with a Tip as with `Equinox.CosmosStore`)
-- the State is established by [`fold`](https://en.wikipedia.org/wiki/Fold_(higher-order_function))ing the sequence of `Event`s, starting from an `initial` state
+- a `State` type, on which decisions can be based. This can be updated as a consequence of a decision. e.g. the item identifiers in a cart and the associated quantities
```
ragiano215, line 859:
```suggestion
- a `State` type, on which decisions can be based. This can be updated as a consequence of a decision, e.g. the item identifiers in a cart and the associated quantities
```

README.md
```diff
@@ -854,12 +856,12 @@ _When applying the concept of a Decider to event sourcing, the consistency requi
...
+- a `decide` function, which is presented a `State`, and returns a decision, which we use to update the State, if, and only if there is no concurrency conflict when applying them. e.g. the `decider` might validate that it's acceptable to start a process at the present time, returning the identifier of the process; if there is already one in flight, it can return the identifier of that already-started process (covered under the term idempotency further onwards)
```
ragiano215, line 860:
```suggestion
- a `decide` function, which is presented a `State`, and returns a decision, which we use to update the State, if, and only if there is no concurrency conflict when applying them, e.g. the `decider` might validate that it's acceptable to start a process at the present time, returning the identifier of the process; if there is already one in flight, it can return the identifier of that already-started process (covered under the term idempotency further onwards)
```

README.md
```diff
@@ -884,18 +886,19 @@ I'd present the fact that Equinox:
 #### ... about the process of making _decisions_
 
 Finally, I'd say that a key thing the Decider concept brings is a richer way of looking at event sourcing than the typical event sourcing 101 examples you might see:
-- de-emphasizing one-way things that map commands to events without deduplicating and/or yielding a result (not saying that you shouldn't do the simplest thing -- you absolutely should)
-- de-emphasizing all projection handlers only ever just sitting there looking for `MyThingCreated` and doing an `INSERT` with a try/catch for duplicate inserts as a way to make it idempotent (and every stream design requiring a Created event as a way to enable that)
+- de-emphasizing one way calls into the void that map commands to events without deduplicating and/or yielding a result (not saying that you shouldn't do the simplest thing -- you absolutely should)
+- de-emphasizing all the notion that all projection handlers dont get more exciting than sitting there looking for `MyThingCreated` and doing an `INSERT` with a try/catch for duplicate inserts as a way to make it idempotent (and that every stream design must have a Created event because that's the magic recipe)
```
ragiano215, line 890:
```suggestion
- de-emphasizing the notion that all projection handlers don't get more exciting than sitting there looking for `MyThingCreated` and doing an `INSERT` with a try/catch for duplicate inserts as a way to make it idempotent (and that every stream design must have a Created event because that's the magic recipe)
```

PR #303, opened 2021-11-18
CosmosStore: Cleanup Cosmos calls

bartelink:
Long overdue cleanup in CosmosStore - initial logic was intentionally matching V2 init logic, but it can be clarified now based on:
- support for Autoscale from @belcher-rok 
- the V3 SDK has Database and Container objects where V2 did not

### DIFF
--- a/src/Equinox.CosmosStore/CosmosStore.fs
+++ b/src/Equinox.CosmosStore/CosmosStore.fs
@@ -571,73 +571,62 @@ module Initialization =
     let (|ThroughputProperties|) = function
         | Throughput.Manual rus -> ThroughputProperties.CreateManualThroughput(rus)
         | Throughput.Autoscale maxRus -> ThroughputProperties.CreateAutoscaleThroughput(maxRus)
-    let (|MaybeThroughputProperties|) = Option.map (function ThroughputProperties tp -> tp) >> Option.toObj
-    let adjustOfferC (c:Container) (ThroughputProperties tp) = async {
-        let! ct = Async.CancellationToken
-        let! _ = c.ReplaceThroughputAsync(tp, cancellationToken = ct) |> Async.AwaitTaskCorrect in () }
-    let adjustOfferD (d : Database) (ThroughputProperties tp) = async {
-        let! ct = Async.CancellationToken
-        let! _ = d.ReplaceThroughputAsync(tp, cancellationToken = ct) |> Async.AwaitTaskCorrect in () }
-    let private createDatabaseIfNotExists (client : CosmosClient) dName (MaybeThroughputProperties tpOrNull) = async {
-        let! ct = Async.CancellationToken
-        let! dbr = client.CreateDatabaseIfNotExistsAsync(dName, throughputProperties = tpOrNull, cancellationToken = ct) |> Async.AwaitTaskCorrect
-        return dbr.Database }
     let private createOrProvisionDatabase (client : CosmosClient) dName mode = async {
-        match mode with
-        | Provisioning.Database throughput ->
-            let! db = createDatabaseIfNotExists client dName (Some throughput)
-            do! adjustOfferD db throughput
-        | Provisioning.Container _ | Provisioning.Serverless ->
-            let! _ = createDatabaseIfNotExists client dName None in () }
-    let private createContainerIfNotExists (d : Database) (cp : ContainerProperties) (MaybeThroughputProperties tpOrNull) = async {
         let! ct = Async.CancellationToken
-        let! c = d.CreateContainerIfNotExistsAsync(cp, throughputProperties = tpOrNull, cancellationToken = ct) |> Async.AwaitTaskCorrect
-        return c.Container }
-    let private createOrProvisionContainer (d : Database) (cp : ContainerProperties) mode = async {
+        let createDatabaseIfNotExists maybeTp = async {
+            let! r = client.CreateDatabaseIfNotExistsAsync(dName, throughputProperties = Option.toObj maybeTp, cancellationToken = ct) |> Async.AwaitTaskCorrect
+            return r.Database }
         match mode with
-        | Provisioning.Container throughput ->
-            let! c = createContainerIfNotExists d cp (Some throughput)
-            do! adjustOfferC c throughput
-            return c
-        | Provisioning.Database _ | Provisioning.Serverless ->
-            return! createContainerIfNotExists d cp None }
+        | Provisioning.Container _ | Provisioning.Serverless -> return! createDatabaseIfNotExists None
+        | Provisioning.Database (ThroughputProperties tp) ->
+            let! d = createDatabaseIfNotExists (Some tp)
+            let! _ = d.ReplaceThroughputAsync(tp, cancellationToken = ct) |> Async.AwaitTaskCorrect
+            return d }
+    let private createOrProvisionContainer (d : Database) (cName, pkPath, customizeContainer) mode = async {
+        let cp = ContainerProperties(id = cName, partitionKeyPath = pkPath)
+        customizeContainer cp
+        let! ct = Async.CancellationToken
+        let createContainerIfNotExists maybeTp = async {
+            let! r = d.CreateContainerIfNotExistsAsync(cp, throughputProperties = Option.toObj maybeTp, cancellationToken = ct) |> Async.AwaitTaskCorrect
+            return r.Container }
+        match mode with
+        | Provisioning.Database _ | Provisioning.Serverless -> return! createContainerIfNotExists None
+        | Provisioning.Container (ThroughputProperties throughput) ->
+            let! c = createContainerIfNotExists (Some throughput)
+            let! _ = c.ReplaceThroughputAsync(throughput, cancellationToken = ct) |> Async.AwaitTaskCorrect
+            return c }
     let private createStoredProcIfNotExists (c : Container) (name, body): Async<float> = async {
-        try let! r = c.Scripts.CreateStoredProcedureAsync(Scripts.StoredProcedureProperties(id = name, body = body)) |> Async.AwaitTaskCorrect
+        let! ct = Async.CancellationToken
+        try let! r = c.Scripts.CreateStoredProcedureAsync(Scripts.StoredProcedureProperties(id = name, body = body), cancellationToken = ct) |> Async.AwaitTaskCorrect
             return r.RequestCharge
-        with (:? Microsoft.Azure.Cosmos.CosmosException as ce) when ce.StatusCode = System.Net.HttpStatusCode.Conflict -> return ce.RequestCharge }
-    let private mkContainerProperties containerName partitionKeyFieldName =
-        ContainerProperties(id = containerName, partitionKeyPath = sprintf "/%s" partitionKeyFieldName)
-    let private createBatchAndTipContainerIfNotExists (client : CosmosClient) (dName, cName) mode : Async<Container> =
-        let def = mkContainerProperties cName Batch.PartitionKeyField
-        def.IndexingPolicy.IndexingMode <- IndexingMode.Consistent
-        def.IndexingPolicy.Automatic <- true
+        with :? CosmosException as ce when ce.StatusCode = System.Net.HttpStatusCode.Conflict -> return ce.RequestCharge }
+    let private applyBatchAndTipContainerProperties (cp : ContainerProperties) =
+        cp.IndexingPolicy.IndexingMode <- IndexingMode.Consistent
+        cp.IndexingPolicy.Automatic <- true
         // Can either do a blacklist or a whitelist
         // Given how long and variable the blacklist would be, we whitelist instead
-        def.IndexingPolicy.ExcludedPaths.Add(ExcludedPath(Path="/*"))
+        cp.IndexingPolicy.ExcludedPaths.Add(ExcludedPath(Path="/*"))
         // NB its critical to index the nominated PartitionKey field defined above or there will be runtime errors
-        for k in Batch.IndexedFields do def.IndexingPolicy.IncludedPaths.Add(IncludedPath(Path = sprintf "/%s/?" k))
-        createOrProvisionContainer (client.GetDatabase dName) def mode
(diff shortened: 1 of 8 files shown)


bartelink:
I couldn't help pulling at the thread ;)
Needs testing before merge and there's no rush

src/Equinox.CosmosStore/CosmosStore.fs
```diff
@@ -394,8 +394,7 @@ module private MicrosoftAzureCosmosWrappers =
     type Container with
         member container.TryReadItem(partitionKey : PartitionKey, documentId : string, ?options : ItemRequestOptions): Async<float * ReadResult<'T>> = async {
             let! ct = Async.CancellationToken
-            let options = defaultArg options null
-            use! rm = async { return! container.ReadItemStreamAsync(documentId, partitionKey, requestOptions = options, cancellationToken = ct) |> Async.AwaitTaskCorrect }
+            use! rm = async { return! container.ReadItemStreamAsync(documentId, partitionKey, requestOptions = Option.toObj options, cancellationToken = ct) |> Async.AwaitTaskCorrect }
```
deviousasti, line 397: Do we need this `async` block at all?

bartelink: Good point - Struggling to think of the remotest justification so applied it. Same construct is in V2 - I think the original code was grabbed from some generic helpers somewhere!

src/Equinox.CosmosStore/CosmosStore.fs
```diff
@@ -384,22 +384,15 @@ module Log =
...
     type Container with
         member container.TryReadItem(partitionKey : PartitionKey, documentId : string, ?options : ItemRequestOptions): Async<float * ReadResult<'T>> = async {
             let! ct = Async.CancellationToken
-            let options = defaultArg options null
-            use! rm = async { return! container.ReadItemStreamAsync(documentId, partitionKey, requestOptions = options, cancellationToken = ct) |> Async.AwaitTaskCorrect }
-            let rc = rm.Headers.GetRequestCharge()
-            if rm.StatusCode = System.Net.HttpStatusCode.NotFound then return rc, NotFound
-            elif rm.StatusCode = System.Net.HttpStatusCode.NotModified then return rc, NotModified
-            else return rc, Found (container.Database.Client.ClientOptions.Serializer.FromStream<'T>(rm.EnsureSuccessStatusCode().Content)) }
+            use! rm = container.ReadItemStreamAsync(documentId, partitionKey, requestOptions = Option.toObj options, cancellationToken = ct) |> Async.AwaitTaskCorrect
```
bartelink, line 391: @jchannon FYI much cleaner now than when I shared it with you...

src/Equinox.CosmosStore/CosmosStore.fs
```diff
@@ -384,22 +384,18 @@ module Log =
...
-    type Headers with
-        member headers.GetRequestCharge () =
-            match headers.TryGetValue "x-ms-request-charge" with
-            | true, charge -> float charge
-            | _ -> 0.
-
     type ReadResult<'T> = Found of 'T | NotFound | NotModified
     type Container with
-        member container.TryReadItem(partitionKey : PartitionKey, documentId : string, ?options : ItemRequestOptions): Async<float * ReadResult<'T>> = async {
+        member private container.DeserializeResponseBody<'T>(rm : ResponseMessage) : 'T =
```
bartelink, line 389: @ealsur could something like this be exposed on the SDK in order to make the deferred deserialization clean and not reliant on spelunking in the source? (Maybe as an extension method on the `ResponseMessage` either?)

bartelink: while I think the code in https://github.com/jet/equinox/blob/eade429557a857fedbaa0227ec28358f90a3bc50/src/Equinox.CosmosStore/CosmosStore.fs#L391-L395 is pretty clear for my context, it's not obvious and the xmldoc does not hint at it being possible 
(also, It's very possible my impl is not optimal, in which case I'd like to be using an SDK method that will be)

ealsur: Sorry for the late response, missed this notification. Stream APIs are by nature unhooked from serialization, so the user can choose whatever serialization engine they want (or simply return the Stream to a caller if you are working in a middleware).

We don't expect users to consume the Stream APIs and also use our own Serializer, in that case, why are you using the Stream API? (sorry, my F# is non-existent, so I might be missing something).

bartelink: This was via https://twitter.com/OliTomli/status/1466055639906590723
Recall non-200s throw on the non-stream API - the code above is derived from V2 SDK code and eventually could/will do its own deserialization (using STJ) but my real broad point is that adding a helper like this or maybe an overload with a callbck would allow one to mix and match the Stream and non-Stream variants when you discover you want to filter out or provide alternate processing for a given status

bartelink:
Ping @ealsur I added some answers to your questions above - I still think there should be some way for the API to provide a ramp between a call that forces deserialization regardless of response code versus requiring the consumer to take on all the burden the minute they are interested in handling non-200 cases

ealsur:
@bartelink Isn't the `ResponseMessage.IsSuccessStatusCode` property and `ResponseMessage.EnsureSuccessStatusCode()` method if you want an exception to be thrown enough to fill that gap currently?

bartelink:
@ealsur Its about avoiding the exception if you are handling a non-200 case. Let me try to explain betrer...

The ask is for something like my `DeserializeResponseBody` helper in:
https://github.com/jet/equinox/blob/tidy-cosmosinit/src/Equinox.CosmosStore/CosmosStore.fs#L389-L391

This enables the conditional reading of the body by first looking at the status code in https://github.com/jet/equinox/blob/tidy-cosmosinit/src/Equinox.CosmosStore/CosmosStore.fs#L392-L398

A pidgin C# translation would be
```
Result TryReadItem<T>(this container, string id, PartitionKey pk) {
    var rm = await container.ReadItemStreamAsync<T>(id, partitionKey, options, ct)
    rm.StatusCode switch
        case System.Net.HttpStatusCode.NotFound: return Result.NotFound
        case System.Net.HttpStatusCode.NotModified: return Result.NotModified
        case default: Result.Found(container.DeserializeResponseBody<T>(rm))
```
The point is you want something pithy that lets you deserialize to a specific type after you have checked the StatusCode on the ResponseMessage ( `container.DeserializeResponseBody<T>(rm)` in my code)

... figuring out that what `ReadItemAsync` does to achieve this is "simply" `container.Database.Client.ClientOptions.Serializer.FromStream<'T>(rm.EnsureSuccessStatusCode().Content)` was not obvious to me.

How exactly to surface it give the serializer us on the container/client and not known to the ResponseMessage is the fun bit

ealsur:
So the ask is how to achieve typed APIs (for example, `ReadItemAsync`) but not throw exceptions?

bartelink:
Yes and no. The ask is:
- do the call, throw on failure etc 
- expose the `StatusCode` etc without triggering the reading/decoding/waiting for the content to get it
- (if I'm still interested after that check) let me await and decode the body

ealsur:
I'm truly sorry but there is a part I don't understand in the sequence, the only scenario where the shared Tweet makes sense is from a performance perspective (throwing causes an overhead). Assuming the operation was a failure and an exception was thrown, there is no body to deserialize. The body is only deserialized on the success path (on the failure really the backend did not send any body).

Let's run a scenario through the proposed steps, for example, a ReadItem returning 404.

1. The server returns a 404, there is no body in the response.
2. The SDK throws a CosmosException, the exception contains the 404 StatusCode.
3. What body would you decode if the serializing is deferred? There is nothing to read.

All the error code situations contain a server response whose body is either empty or might contain some extra information of the error itself (case of TransactionalBatch), but nothing that it's useful for the original operation.

If your intent is simply to defer serialization only on success cases, then that's what the SDK does already.
If your intent is to not have exceptions (what the Tweet was asking) then I don't think that's a behavior that can change in the SDK and the only "workaround" would be something like the proposed snippet.

bartelink:
The primary ask is to avoid exceptions - I fully expect and am ready to handle 304. It's not Exceptional. Its like Parse vs TryParse except we want diagnostics for both success and fail cases, not just a `bool`

As it happens, for my use cases, I want/need the body in all success cases.

If one cannot expect the body to be present, then I don't want to read it. 

Dropping to the Stream API happens to make it possible to do what I need, but in the most abstract sense, the need is some variant of `ReadItemAsync` that allows one to decouple the two steps that are bundled in it at present:
a) do the work; expose the result, _BUT don't throw if it is not "Successful"; I want to do my own interpretation, esp for 304_
b) let me deserialize the body (feel free throw if I neglected to guard that call with a check for a StatusCode that would imply one should be available)

ealsur:
I see, so basically the ask is an API to invoke the internal serializer with a ResponseMessage response? For point operations it would be simple, but the real issue is for query responses or other feedtype responses. 

For example, if you have a ResponseMessage that represents a ReadItem, you could call `whatever.DeserializeBody<T>(ResponseMessage)` where `T` is your type intended for the ReadItem. But if the ResponseMessage is coming from a ReadNextAsync from a query, then a generic API like `whatever.DeserializeBody<T>(ResponseMessage)` would be prompt to failure. The problem is the input can be anything.

@kirankumarkolli / @j82w - Do you folks have any thoughts on this?

bartelink:
Yes, that represents the ask well.

You raise good points about this only applying for individual `[Try]ReadItemAsync` Response Messages, and I would definitely agree that putting it in the wrong place could make a mess that no amount of remarks in xmldoc can save people from.

Perhaps a "V4"-timescale revised API design can expose it in a way that is discoverable and covers all the cases. ... an, in the interim the `container.Database.Client.ClientOptions.Serializer.FromStream<'T>(rm.EnsureSuccessStatusCode().Content)` code snippet can live in the issue that documents the requirement until it's all resolved ;)

j82w:
Would adding an operation type to the ResponseMessage allow it to be handled correctly?

kirankumarkolli:
ReadItemAsync<T> is a wrapper around stream API, doing exactly the two steps mentioned in https://github.com/jet/equinox/pull/303#issuecomment-1006727754 

It's very simple and straight forward. As in most cases having it as an extension method like ReadItemAsync404NotThrow() on Container also might make it more idiomatic.

bartelink:
>It's very simple and straight forward. As in most cases having it as an extension method like ReadItemAsync404NotThrow() on Container also might make it more idiomatic.

I think the problem here is that, as in my case, you frequently want to handle 304 etc in addition to 404, and having a custom extension for each case won't solve that, and you need to epose the ResponseMessage at some point in order to let people pull out the `RequestCharge` etc.

So for me, any real solution needs to do one of two things:
1. expose an equivalent of `container.DeserializeResponseBody<T>(rm)`
2. have the caller pass a function that can do the equivalent via a `Func` passed to it

For example, if there was a:

```
TResult TryReadItem<TResult,TBody>(this container, string id, PartitionKey pk, Func<ResponseMessage, Func<TBody>> interpret, CancellationToken ct);
```
Then Mathias's example can become:
```
Result ReadOrNotFound<T>(this container, string id, PartitionKey pk, ct) =>
    container.TryReadItem(id, pk, cancellationToken = ct, interpret = (rm, readBody) =>
        rm.IsSuccessStatusCode
            ? Result.Found(readBody())
            : Result.Failed(rm.StatusCode));
```
And mine becomes:
```
        member container.TryReadItem(partitionKey : PartitionKey, id : string, ?options : ItemRequestOptions): Async<float * ReadResult<'T>> = async {
            let! ct = Async.CancellationToken
            let extractResult (rm : ResponseMessage, readBody : unit -> 'T) =
                rm.Headers.RequestCharge, // NOTE I want the request charge in all cases
                match rm.StatusCode with
                | System.Net.HttpStatusCode.NotFound -> NotFound
                | System.Net.HttpStatusCode.NotModified -> NotModified
                | _ -> readBody()|> Found }
            return! container.TryReadItem(id, partitionKey, requestOptions = Option.toObj options, extractResult, cancellationToken = ct) |> Async.AwaitTaskCorrect
```
While that also solves the problem, the question is how you feel about a `Func` that passes a `Func`, and the fact that there are two generic types involved in the signature.

ealsur:
Just a comment, 304 does never happen on ReadItem, the only scenario 304 is involved is for Change Feed operations.

bartelink:
> Just a comment, 304 does never happen on ReadItem, the only scenario 304 is involved is for Change Feed operations.

I'm over 99% sure that's not true - Equinox makes extensive use of conditional gets with etags to avoid paying for retrieval of bodies that have not change

You can see the condition being applied [in this file](https://github.com/jet/equinox/blob/tidy-cosmosinit/src/Equinox.CosmosStore/CosmosStore.fs#L615)

ealsur:
> I'm over 99% sure that's not true - Equinox makes extensive use of conditional gets with etags to avoid paying for retrieval of bodies that have not change
> 
> You can see the condition being applied [in this file](https://github.com/jet/equinox/blob/tidy-cosmosinit/src/Equinox.CosmosStore/CosmosStore.fs#L640) (Amusingly, I've just spotted I was erroneously labelling and logging as it as 302...)

This is super interesting, I was not aware a Read Item could return 304 using ETag (I was under the assumption that ETag was mainly used for optimistic concurrency), and could not find this use case in our documentation either, thanks for sharing.

Having a generic deserializing method in my opinion will bring a deal of issues due to mis-use, there is no amount of documentation we can write that will avoid it.

bartelink:
>Having a generic deserializing method in my opinion will bring a deal of issues due to mis-use, there is no amount of documentation we can write that will avoid it.

Yeah, I can't ague with that.

If [an API with a callback to extract from the ResponseMessage](https://github.com/jet/equinox/pull/303#issuecomment-1007229736) was to be provided, I'd definitely use it, but I'll definitely understand if that's not done now ... as long as there is an intent to provide such a facility in the V4 timeframe.

It seems at this point that its reasonable to park this discussion for now with the following conclusions:
- `ReadItemStreamAsync` + `container.Database.Client.ClientOptions.Serializer.FromStream<'T>(rm.EnsureSuccessStatusCode().Content)` is the unofficial way to handle expected 304/404 values without exceptions for the moment
- It would be good to have an issue noting this shortcoming of the API logged on the repo, and that workaround can be mentioned there

PR #304, opened 2021-12-12
Removed Deprecated Grafana Spec

ragiano215:
Removed Grafana spec now that the canonical one is living in the [`dotnet-templates` repo](https://github.com/jet/dotnet-templates). The latter is also more appropriate since it encompasses both Equinox and Propulsion metrics.

(Will be merged when [jet/dotnet-templates/pull/108](https://github.com/jet/dotnet-templates/pull/108) is merged.)


bartelink:
Thanks!

PR #318, opened 2022-03-09
Moving Caching logic to central file

bartelink:
Prep/cleanup in advance of #196 potentially introducing a third clone of the caching logic into the codebase

### DIFF
--- a/src/Equinox.Core/Cache.fs
+++ b/src/Equinox.Core/Cache.fs
@@ -24,52 +24,6 @@ type ICache =
     abstract member UpdateIfNewer : key: string * options: CacheItemOptions * entry: CacheEntry<'state> -> unit
     abstract member TryGet : key: string -> (StreamToken * 'state) option
 
-module Caching =
-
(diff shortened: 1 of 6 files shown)


bartelink:
@DSilence Any idea whether there's a running system out there that still relies on it? I'm thinking of removing it (i.e. this PR) 4.x unless this actually got used

bartelink:
Caching.fs will move into EventstoreDb and be linked into EventStore and SqlStreamStore rather than living in Equinox.Core (see #196)

DSilence:
Not sure there is anything in prod currently relying on async implementation. Could be useful for file-based cache though (or anything network related)

bartelink:
@DSilence hard to argue against that reasoning (even if speculative!); thanks for the nudge; reverted in [6be1b5a](https://github.com/jet/equinox/pull/318/commits/6be1b5a5528fa71c8a68af1bbde6406d7475c79c)
I guess it also makes for there to be asynchrony here if #61 ever happens, given that's definitely Async in nature

PR #328, opened 2022-05-26
CosmosStore - Extend aux collection initialization

brihadish:
Modifies initialization of aux collection to take an explicit input representing the cosmos provisioning mode. This change will be used in 'Propulsion.Tool' for supporting autoscale option and 'serverless' cosmos database.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ The `Unreleased` section name is replaced by the expected version of next releas
 - `Equinox.Tool`/`samples/`: switched to use `Equinox.EventStoreDb` [#196](https://github.com/jet/equinox/pull/196)
 - Update all non-Client dependencies except `FSharp.Core`, `FSharp.Control.AsyncSeq` [#310](https://github.com/jet/equinox/pull/310)
 - Update all Stores to use `FsCodec` v `3.0.0`, with [`EventBody` types switching from `byte[]` to `ReadOnlyMemory<byte>`, see FsCodec#75](https://github.com/jet/FsCodec/pull/75) [#323](https://github.com/jet/equinox/pull/323)
+- `CosmosStore`: Replace the hard-coding in `Equinox.CosmosStore.Core.Initialization.initAux` with an input parameter [#328](https://github.com/jet/equinox/pull/328) 
 
 ### Removed
  
--- a/src/Equinox.CosmosStore/CosmosStore.fs
+++ b/src/Equinox.CosmosStore/CosmosStore.fs
@@ -581,7 +581,7 @@ module Initialization =
         cp.IndexingPolicy.Automatic <- false
         cp.IndexingPolicy.IndexingMode <- IndexingMode.None
 
-    let initAux (client : CosmosClient) (dName, cName) rus mode = async {
+    let initAux (client : CosmosClient) (dName, cName) mode = async {
         let! d = createOrProvisionDatabase client dName mode
         return! createOrProvisionContainer d (cName, "/id", applyAuxContainerProperties) mode } // as per Cosmos team, Partition Key must be "/id"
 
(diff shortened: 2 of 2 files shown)


bartelink:
Now is definitely a reasonable time to introduce a breaking change like this given that we are in a `-beta` phase.

I have a question though... Can you share the configuration you've arrived at and the reasoning behind that? It will help others understand, and perhaps it might reveal a better solution.

I'm slightly concerned that e.g. autoscale being too aggressive would cause lease drop and reassignment storms.

The other thing is that appropriate wiring should probably follow to surface the facility in [Propulsion.Tool](https://github.com/jet/propulsion/blob/fafee4f374ec6346f2be2df79d910bf1c99ac807/tools/Propulsion.Tool/Program.fs#L92)

My aim would be that someone randomly running that tool would probably end up with the prev manual provisioning of 400, and not blindly go in with a more exotic configuration without understanding the implications. The dream would be to have a section in Propulsion's DOCUMENTATION.md with an overview and/or links to guidance on the topic.

An example for me of the subtleties involved: https://github.com/Azure/azure-cosmos-dotnet-v3/issues/3215#issuecomment-1136012525

The relevant bits in Equinox.Tool to borrow would be
- https://github.com/jet/equinox/blob/54b373b98ee9370daee6b2a54ba91fca087d09b0/tools/Equinox.Tool/Program.fs#L44-L68
- https://github.com/jet/equinox/blob/54b373b98ee9370daee6b2a54ba91fca087d09b0/tools/Equinox.Tool/Program.fs#L367-L384

You should be able to do a draft PR by changing the `PackageReference` for `Equinox.CosmosStore` over there to a local `ProjectReference` 

----

Having said that, if you have a definite need for this and intend to simply call this API yourself directly, I definitely don't have an objection to the removal of the hardcoding of a default; my default stance is to push all these concerns outward as you are doing, and I agree there is no value in Equinox being the lib to take an opinionated stance on something as subtle as this which is ultimately a nuanced thing.

Assuming you wish to go ahead with the PR, can you also add a changelog.md entry please?

brihadish:
Thanks for the review :).

The reasoning behind this PR is that we have been getting compliance requirements from our DevOps team to move to 'serverless' dbs for non-prod environments and for universally applying 'autoscale' setting in collections pertaining to 'provisioned throughput' dbs. The aux collection was preventing us from undertaking an endeavor meet the aforementioned compliance requirements.

I do agree that a setting 400 RUs for the aux collection should be more than sufficient and I have already done the necessary changes to Propulsion.Tool which exactly matches your proposal (Default behavior of manual 400 RUs for backward compatibility. 'Autoscale' option will use a default max RU of 4000 RUs so that the fixed cost is same as manual mode.) I was unsure how to push the Propulsion tool changes without an updated nuget for 'Equinox.CosmosStore'. 

Please correct if I'm wrong. I think using a local 'ProjectReference' will include bringing multiple projects into the propulsion PR. Is that fine? Nevertheless, I will create a draft PR of propulsion tool without the reference fix. I will also make necessary changes to the propulsion tool documentation.

I will certainly make an entry in changelog.md and update the current PR.

bartelink:
>I was unsure how to push the Propulsion tool changes without an updated nuget for 'Equinox.CosmosStore'.
You are correct that it's not possible to have it build on CI without putting lots of source in the PR branch etc, but you can show the code as intended and get the bulk of the view out of the way.

So what I meant is for you to do it locally, push the Propulsion bits and leave the PR in Draft state on github.

>Please correct if I'm wrong. I think using a local 'ProjectReference' will include bringing multiple projects into the propulsion PR. Is that fine? Nevertheless, I will create a draft PR of propulsion tool without the reference fix. I will also make necessary changes to the propulsion tool documentation.

For it to work happily locally you would need to add  ../equinox/src/Equinox.CosmosStore to the sln (I think it will be OK without Equinox.Core and Equinox)

Any docs piece would be wonderful.

---
Once you have the changelog bit pushed, I can stage a release of Equinox which you can target once it gets to nuget.

Is having it in Equinox.CosmosStore 4.0.0-beta.X and your changes being in Propulsion.Tool-2.13.0-beta.Y OK ?

(I dont any significant changes in Propulsion or Equinox but I'll be on vacation for a period and would prefer to not make badge them with full released numbers until I'm ready to do the associated changes for `dotnet-templates`)

brihadish:
Propulsion tool PR is https://github.com/jet/propulsion/pull/142. Using beta packages for the propulsion tool will be fine.

bartelink:
Thanks for your contribution! Will ping you here when the nuget upload has completed (there's a manual step in the chain; no eta as yet)

deviousasti:
Packages released as [beta6](https://www.nuget.org/packages/Equinox/4.0.0-beta.6).

PR #344, opened 2022-11-11
flip argument order for context taking functions putting the context last

nordfjord:
This flips the argument order from `context log cat` to `log cat context`

This makes it easier to switch from not using a context to using a context since the APIs are equivalent

```diff
- Category(....) |> Decider.resolve log
+ Category(....) |> Decider.resolveWithContext log
```

### DIFF
--- a/src/Equinox.Core/Category.fs
+++ b/src/Equinox.Core/Category.fs
@@ -70,9 +70,9 @@ module Decider =
 type DeciderExtensions =
 
     [<System.Runtime.CompilerServices.Extension>]
-    static member Resolve(log, cat : Category<'event, 'state, 'context>, context) : System.Func<struct (string * string), DeciderCore<'event, 'state>> =
+    static member Resolve(cat : Category<'event, 'state, 'context>, log, context) : System.Func<struct (string * string), DeciderCore<'event, 'state>> =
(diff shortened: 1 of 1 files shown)


bartelink:
Thank you! As discussed on the discord, this make a lot of sense; thanks for taking the time to convince me of that.
I still don't love the `resolveWithContext` name - it's cumbersome and context is pretty overloaded
Any suggestions and followups in this space are definitely welcomed

nordfjord:
Okay... Hear me out... What if we just got rid of the unit hiding methods?

bartelink:
> Okay... Hear me out... What if we just got rid of the unit hiding methods?

Hm
1. that'd lose consistency with the C# resolve not asking for a `unit`
2. I'd have to see where the random parens land a) in some representative code b) in the samples and tests in this repo

I think I spiked it and didnt like it, but I didnt have them in this order at the time and didnt have all these baroquely named helper methods

The bottom line is that this is a pretty advanced feature and will likely remain so (even though working without it will probably be like not using OTel and soon enough) - as such the cost of it making the syntax wierd for someone relatively new to both F# and Equinox needs to be weighed carefully...

bartelink:
Thanks! Merging for now; followups welcome

PR #346, opened 2022-11-11
Proof of concept: use Activity rather than serilog

nordfjord:


tests/Equinox.EventStore.Integration/Infrastructure.fs
```diff
@@ -66,6 +65,26 @@ type LogCapture() =
...
+            | "ReadLast"  -> EsAct.ReadLast |> captures.Add
+            | "Write"     -> captures.Add(if act.Status = ActivityStatusCode.Error then EsAct.AppendConflict else EsAct.Append)
+            | _ -> ()
+        ActivitySource.AddActivityListener(l)
+        l
+
+    member _.ExternalCalls = captures |> List.ofSeq
+    member _.Clear() = captures.Clear()
+    interface IDisposable with
+        member this.Dispose() = (listener :> IDisposable).Dispose()
```
nordfjord, line 87: Attempting to mirror the API for the serilog version here.

samples/Web/SerilogExporter.fs
```diff
@@ -0,0 +1,24 @@
...
+
+  let log = LoggerConfiguration()
+                .WriteTo.Console(
+                    outputTemplate = "[{Timestamp:HH:mm:ss} {Level:u3}] {Message:lj} {Properties}{NewLine}{Exception}",
+                    theme = Sinks.SystemConsole.Themes.AnsiConsoleTheme.Code).CreateLogger() :> ILogger
+
+  override this.Export(batch) =
+      for span in batch do
+          let mutable logger = log
+          for tag in span.TagObjects do
```
bartelink, line 18: this was a job for.... foldman!

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -14,229 +16,127 @@ type EventBody = ReadOnlyMemory<byte>
...
-                        | MetricEvent (Read stats) -> LogSink.Read.Ingest stats
-                        | MetricEvent (ReadL stats) -> LogSink.ReadL.Ingest stats
-                        | MetricEvent (Write stats) -> LogSink.Write.Ingest stats
-                        | MetricEvent (Resync stats) -> LogSink.Resync.Ingest stats
-                        | MetricEvent (Rollup _) -> ()
-                        | _ -> ()
-
-        /// Relies on feeding of metrics from Log through to Stats.LogSink
-        /// Use Stats.LogSink.Restart() to reset the start point (and stats) where relevant
-        let dump (log : ILogger) =
```
bartelink, line 105: example output of this in-context (TL;DR need to be able to produce similar stuff to that so one can analyze perf in the context of of an app that's saturating a rate limited store)

Every minute the scheduler dumps throughput info, and the equinox store module logs the summary of what it did (the same Metrics get fed to prometheus too)

image

bartelink: Not sure what to name it that's better but I'll grant you it's not something one would just discover! 

Yes, totally agree proper o11y stack is the true answer, but don't underestimate this poor man's aggregation mechanism can work very well for 90% of stuff - even when having the grafana charts rigged, for most ingestion/reaction jobs I tend to glance at the high level breakdown above (yes, despite it only having a pathetic average!) rather than the chart.

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -88,13 +84,13 @@ module private Read =
             span
                 .AddTag("eqx.stream_name", streamName)
```
bartelink, line 85: suggest `let inline prop k v = (x:Span) = x.AddTag(k, v)` helper a la `Log.prop` to 1) remove parens and commas 2) let formatting stay almost the same
then also do partially applied ones, e.g. `let propSN = prop "eqx.streamName`

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -88,13 +84,13 @@ module private Read =
...
-                .AddTag("eqx.start_post", startPos)
-                .AddTag("eqx.requires_leader", requiresLeader)
-                .AddTag("eqx.attempt", attempt)
+                .AddTag("eqx.start_pos", startPos)
                 .AddTag("eqx.version", slice.LastVersion)
                 .AddTag("eqx.bytes", bytes)
-                .AddTag("eqx.count", count)
-            |> ignore
+                .AddTag("eqx.count", count) |> ignore
+            if requiresLeader then span.AddTag("eqx.requires_leader", requiresLeader) |> ignore
```
bartelink, line 91: `let propLeader v = if v then prop "eqx.requires_leader` else ignore`

nordfjord:
Closing as we've got most of the benefits down in #354

PR #347, opened 2022-11-12
Write out the loop for better perf on chooseV

nordfjord:
See a comparioson of the ASM [here](https://sharplab.io/#v2:DYLgZgzgNALiCGEC2UAmIDUAfJB7VArsAKYAEAKgBYCWAdgOYCwAUCyTKQMaW64TEA1UgAowIUnQ4BaAHwTaHAG64ADjGq5aAShHwATnvGTS+vfACeOgLykA2llJhceiTGJJ5Jg6VS5SSeBhuR1d3UgB3ahhKUgF4YAJiAGVcJDJFUllSc2piYFRSDIc4hOIAOU0yLOEdLABdFjZiDhhcAHk1DVpSGzACWk51TVj4xJS0wsy5cfTSYtHyyqnSCtpiJo5uXn4BACYRMXlpOWNlTs1rUgBBAwsAOi2+MlFSGTlWjqHtIA=)

the new version doesn't need to allocate an option for each voption either.

Here's the result from a Benchmark.NET run as well

|     Method |     Mean |    Error |   StdDev |   Gen0 | Allocated |
|----------- |---------:|---------:|---------:|-------:|----------:|
| ChooseVOld | 48.85 ns | 0.167 ns | 0.148 ns | 0.0162 |     136 B |
| ChooseVNew | 32.74 ns | 0.725 ns | 1.923 ns | 0.0076 |      64 B |

The old method allocates a new array and options, the new one only allocates the array

### DIFF
--- a/src/Equinox.Core/Infrastructure.fs
+++ b/src/Equinox.Core/Infrastructure.fs
@@ -71,4 +71,4 @@ module Seq =
 
 module Array =
 
-    let inline chooseV f = Array.choose (f >> ValueOption.toOption)
+    let inline chooseV f arr = [| for item in arr do match f item with ValueSome v -> yield v | ValueNone -> () |]
--- a/src/Equinox.Core/Infrastructure.fs
+++ b/src/Equinox.Core/Infrastructure.fs
@@ -60,15 +60,10 @@ module ValueTuple =
     let inline fst struct (f, _s) = f
     let inline snd struct (_f, s) = s
 
-module ValueOption =
-
-    let inline toOption x = match x with ValueSome x -> Some x | ValueNone -> None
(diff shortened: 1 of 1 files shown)


bartelink (commented):
Cool, that's very welcome - I definitely held my nose as I wrote that code and am not in the least surprised it's disgusting; I was even doing perf tuning at the time, but never tune stuff I can't get to the top of a profile!
Can you do it for the Seq.choose too, and remove the ValueOption.toOption impl please (it seems unused)

nordfjord:
The toArray was unused as well so I've removed the entire ValueOption module

bartelink:
Nice work, thanks!

PR #352, opened 2022-11-15
MessageDB polish and fixes

nordfjord:
I did some debugging and found that Npgsql loads the chunk in memory when doing `ExecuteReadAsync` so there's no reason to use the `ReadAsync` method. The behavior now mirrors what [Npgsql.Fsharp](https://github.com/Zaid-Ajaj/Npgsql.FSharp/blob/master/src/Npgsql.FSharp.fs#L455) does

Then I also noticed that a string was being allocated by npgsql for the data and meta fields so the getting of the field as a byte[] wasn't providing any benefit. I've made the change to UTF8 bytes explicit.

It also turns out that postgres can store nullable nulls 😂 

```sql
create table if not exists test (id int not null, json jsonb, primary key (id));
insert into test(id, json) values (1, null);
insert into test(id, json) values (2, 'null');
select * from test;
 id | json
----+------
  1 |
  2 | null
(2 rows)

insert into test(id, json) values (3, '');
ERROR:  invalid input syntax for type json
LINE 1: insert into test(id, json) values (3, '');
                                              ^
DETAIL:  The input string ended unexpectedly.
CONTEXT:  JSON data, line 1:
```

### DIFF
--- a/src/Equinox.MessageDb/MessageDbClient.fs
+++ b/src/Equinox.MessageDb/MessageDbClient.fs
@@ -80,8 +80,7 @@ type MessageDbReader internal (connectionString : string, leaderConnectionString
         cmd.Parameters.AddWithValue("StreamName", NpgsqlDbType.Text, streamName) |> ignore
         use! reader = cmd.ExecuteReaderAsync(ct)
 
-        let! hasRow = reader.ReadAsync(ct)
-        if hasRow then return [| parseRow reader |]
+        if reader.Read() then return [| parseRow reader |]
         else return Array.empty }
 
     member _.ReadStream(streamName : string, fromPosition : int64, batchSize : int64, requiresLeader, ct) = task {
@@ -99,10 +98,5 @@ type MessageDbReader internal (connectionString : string, leaderConnectionString
         use! reader = cmd.ExecuteReaderAsync(ct)
 
         let events = ResizeArray()
-        let! hasRow = reader.ReadAsync(ct)
-        let mutable hasRow = hasRow
-        while hasRow do
-            events.Add(parseRow reader)
-            let! nextHasRow = reader.ReadAsync(ct)
-            hasRow <- nextHasRow
+        while reader.Read() do events.Add(parseRow reader)
         return events.ToArray() }
--- a/src/Equinox.MessageDb/MessageDbClient.fs
+++ b/src/Equinox.MessageDb/MessageDbClient.fs
@@ -104,7 +104,4 @@ type MessageDbReader internal (connectionString : string, leaderConnectionString
         cmd.Parameters.AddWithValue("BatchSize", NpgsqlDbType.Bigint, batchSize) |> ignore
         use! reader = cmd.ExecuteReaderAsync(ct)
 
-        let events = ResizeArray()
-        while reader.Read() do
-            events.Add(parseRow reader)
-        return events.ToArray() }
+        return [| while reader.Read() do parseRow reader |] }
--- a/src/Equinox.MessageDb/MessageDbClient.fs
+++ b/src/Equinox.MessageDb/MessageDbClient.fs
@@ -14,8 +14,17 @@ type MdbSyncResult = Written of int64 | ConflictUnknown
 type private Format = ReadOnlyMemory<byte>
 
 module private Json =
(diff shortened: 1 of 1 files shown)


src/Equinox.MessageDb/MessageDbClient.fs
```diff
@@ -14,8 +14,17 @@ type MdbSyncResult = Written of int64 | ConflictUnknown
...
-  let private jsonNull = JsonSerializer.SerializeToUtf8Bytes(null)
-  let toArray (m: Format) = if m.IsEmpty then jsonNull else m.ToArray()
+  let private jsonNull = ReadOnlyMemory(JsonSerializer.SerializeToUtf8Bytes(null))
+
+  let parseRow idx (reader: DbDataReader) =
+      if reader.IsDBNull(idx) then jsonNull
+      else reader.GetString(idx) |> Text.Encoding.UTF8.GetBytes |> ReadOnlyMemory
+
+  let addParameter (name: string) (value: Format) (p: NpgsqlParameterCollection) =
+      if value.Length = 0 then
```
bartelink, line 24: Nice (but it looks so much more logical as 2 lines IMO!)

src/Equinox.MessageDb/MessageDbClient.fs
```diff
@@ -14,8 +14,17 @@ type MdbSyncResult = Written of int64 | ConflictUnknown
 type private Format = ReadOnlyMemory<byte>
 
 module private Json =
-  let private jsonNull = JsonSerializer.SerializeToUtf8Bytes(null)
-  let toArray (m: Format) = if m.IsEmpty then jsonNull else m.ToArray()
+  let private jsonNull = ReadOnlyMemory(JsonSerializer.SerializeToUtf8Bytes(null))
+
+  let parseRow idx (reader: DbDataReader) =
```
bartelink, line 19: parse (or Get) - (parseRow made me think it was parsing the whole row)

src/Equinox.MessageDb/MessageDbClient.fs
```diff
@@ -99,10 +110,5 @@ type MessageDbReader internal (connectionString : string, leaderConnectionString
         use! reader = cmd.ExecuteReaderAsync(ct)
 
         let events = ResizeArray()
-        let! hasRow = reader.ReadAsync(ct)
-        let mutable hasRow = hasRow
-        while hasRow do
-            events.Add(parseRow reader)
-            let! nextHasRow = reader.ReadAsync(ct)
-            hasRow <- nextHasRow
+        while reader.Read() do events.Add(parseRow reader)
```
bartelink, line 113: 2 lines?

src/Equinox.MessageDb/MessageDbClient.fs
```diff
@@ -99,10 +110,5 @@ type MessageDbReader internal (connectionString : string, leaderConnectionString
         use! reader = cmd.ExecuteReaderAsync(ct)
 
         let events = ResizeArray()
```
bartelink, line 112: if doing overkill, wonder does reader offer length internally before you Read to the end so that could be passed as the Capacity (or maybe use batchSize so ResizeArray would never need to realloc)

Or you could skip worrying about it and just `return [| while reader.Read() -> parseRow reader |]`

src/Equinox.MessageDb/MessageDbClient.fs
nordfjord: I went with from reader `fromReader`

src/Equinox.MessageDb/MessageDbClient.fs
nordfjord: > or maybe use batchSize so ResizeArray would never need to realloc

I think the general case is < 10 events in the stream so I think allocating a whole batch might be counterproductive allocation wise 🤔

nordfjord: I think I'll keep as a resize array for now. I'm not sure what the implications are of a while loop array expression 😅

src/Equinox.MessageDb/MessageDbClient.fs
nordfjord: Yes

PR #356, opened 2022-11-20
add snapshotting to message-db

nordfjord:
As discussed, this is likely not worth the trouble.

I'm sharing it here anyway, I'm not sure whether this should ultimately be offered up as functionality of the messagedb store or not 🤔. I see the MessageDb store as a starting point that teams could eventually grow out of if their scale demands. It provides a decent set of trade-offs for medium throughput systems, and as such adding snapshots should in most cases be unnecessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ _If you're looking to learn more about and/or discuss Event Sourcing and it's my
 - [Amazon Dynamo DB](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Introduction.html): Shares most features with `Equinox.CosmosStore` ([from which it was ported in #321](https://github.com/jet/equinox/pull/321)). See above for detailed comparison.
 - [Azure Cosmos DB](https://docs.microsoft.com/en-us/azure/cosmos-db): contains some fragments of code dating back to 2016, however [the storage model](DOCUMENTATION.md#Cosmos-Storage-Model) was arrived at based on intensive benchmarking (squash-merged in [#42](https://github.com/jet/equinox/pull/42)). The V2 and V3 release lines are being used in production systems. (The V3 release provides support for significantly more efficient packing of events ([storing events in the 'Tip'](https://github.com/jet/equinox/pull/251))).
 - [EventStoreDB](https://eventstore.org/): this codebase itself has been in production since 2017 (see commit history), with key elements dating back to approx 2016. Current versions require EventStoreDB Server editions `21.10` or later, and communicate over the modern gRPC interface.
-- [message-db](http://docs.eventide-project.org/user-guide/message-db): bindings for the `message-db` Postgres event storage system. [See `message-db` docs](http://docs.eventide-project.org/user-guide/message-db). :pray: [@nordfjord](https://github.com/nordfjord)
+- [MessageDB](http://docs.eventide-project.org/user-guide/message-db): bindings for the `message-db` Postgres event storage system. [See `MessageDB` docs](http://docs.eventide-project.org/user-guide/message-db). :pray: [@nordfjord](https://github.com/nordfjord)
 - [SqlStreamStore](https://github.com/SQLStreamStore/SQLStreamStore): bindings for the powerful and widely used (**but presently unmaintained**) SQL-backed Event Storage system, derived from the EventStoreDB adapter. [See SqlStreamStore docs](https://sqlstreamstore.readthedocs.io/en/latest/#introduction). :pray: [@rajivhost](https://github.com/rajivhost)
 
   __NOTE: The underlying `SqlStreamStore` project is presently unmaintained; as such its hard to recommend using it for production scenarios__
@@ -133,7 +133,7 @@ _If you're looking to learn more about and/or discuss Event Sourcing and it's my
 
   - For MySql, it's pretty much the same as for SQL Server, with the proviso that it's likely that it's had (and has) significantly lower adoption and/or proper scrutiny.
   
-  - For Postgres, _`Equinox.MessageDb`_ is a better choice as the underlying [message-db](http://docs.eventide-project.org/user-guide/message-db/) project is actively maintained and has significantly better documentation than SqlStreamStore. The other thing to point out is that `Equinox.SqlStreamStore.Postgres` does not depend on the final version of `SqlStreamStore.Postgres` as there are no equivalent releases of the `.Mysql` and `.SqlServer` variants.
+  - For Postgres, _`Equinox.MessageDb`_ is a better choice as the underlying [MessageDB](http://docs.eventide-project.org/user-guide/message-db/) project is actively maintained and has significantly better documentation than SqlStreamStore. The other thing to point out is that `Equinox.SqlStreamStore.Postgres` does not depend on the final version of `SqlStreamStore.Postgres` as there are no equivalent releases of the `.Mysql` and `.SqlServer` variants.
 
 # Components
 
@@ -184,7 +184,7 @@ Equinox does not focus on projection logic - each store brings its own strengths
 - `Propulsion.DynamoStore.Lambda` [![Propulsion.DynamoStore.Lambda NuGet](https://img.shields.io/nuget/v/Propulsion.DynamoStore.Lambda.svg)](https://www.nuget.org/packages/Propulsion.DynamoStore.Lambda/): Indexes events written by `Equinox.DynamoStore` via a DynamoDB Streams-triggered Lambda. (Depends on `Propulsion.DynamoStore`)
 - `Propulsion.EventStore` [![Propulsion.EventStore NuGet](https://img.shields.io/nuget/v/Propulsion.EventStore.svg)](https://www.nuget.org/packages/Propulsion.EventStore/) Used in the [`propulsion project es`](dotnet-tool-provisioning--benchmarking-tool) tool command; see [`dotnet new proSync` to generate a sample app](#quickstart) using it. ([depends](https://www.fuget.org/packages/Propulsion.EventStore) on `Equinox.EventStore`)
 - `Propulsion.EventStoreDb` [![Propulsion.EventStoreDb NuGet](https://img.shields.io/nuget/v/Propulsion.EventStoreDb.svg)](https://www.nuget.org/packages/Propulsion.EventStoreDb/) Consumes from `EventStoreDB` v `21.10` or later using the gRPC interface. ([depends](https://www.fuget.org/packages/Propulsion.EventStoreDb) on `Equinox.EventStoreDb`)
-- `Propulsion.MessageDb` [![Propulsion.MessageDb NuGet](https://img.shields.io/nuget/v/Propulsion.MessageDb.svg)](https://www.nuget.org/packages/Propulsion.MessageDb) Consumes from a `message-db` store (in Postgres) using `Npgsql`. ([depends](https://www.fuget.org/packages/Propulsion.MessageDb) on `Npgsql`, `Propulsion.Feed`)
+- `Propulsion.MessageDb` [![Propulsion.MessageDb NuGet](https://img.shields.io/nuget/v/Propulsion.MessageDb.svg)](https://www.nuget.org/packages/Propulsion.MessageDb) Consumes from a `MessageDB` store (in Postgres) using `Npgsql`. ([depends](https://www.fuget.org/packages/Propulsion.MessageDb) on `Npgsql`, `Propulsion.Feed`)
 - `Propulsion.Kafka` [![Propulsion.Kafka NuGet](https://img.shields.io/nuget/v/Propulsion.Kafka.svg)](https://www.nuget.org/packages/Propulsion.Kafka/): Provides a canonical `RenderedSpan` that can be used as a default format when projecting events via e.g. the Producer/Consumer pair in `dotnet new proProjector -k; dotnet new proConsumer`. ([depends](https://www.fuget.org/packages/Propulsion.Kafka) on `Newtonsoft.Json >= 11.0.2`, `Propulsion`, `FsKafka`)
 
 ## `dotnet tool` provisioning / benchmarking tool
@@ -513,7 +513,20 @@ MessageDb support is provided in the samples and the `eqx` tool:
 - being able to supply `mdb` flag to `eqx dump`, e.g. `eqx dump -CU "Favorites-ab25cc9f24464d39939000aeb37ea11a" mdb -c "pgconnectionstring"`
 - being able to supply `mdb` flag to Web sample, e.g. `dotnet run --project samples/Web/ -- mdb -c "pgconnectionstring"`
 
-Equinox does not provide utilities for configuring or installing message-db. See [MessageDB's installation documentation](http://docs.eventide-project.org/user-guide/message-db/install.html).
+Equinox does not provide utilities for configuring or installing MessageDB. See [MessageDB's installation documentation](http://docs.eventide-project.org/user-guide/message-db/install.html).
+
+In addition to the default access strategy of reading the whole stream forwards in batches, the following access strategies are supported in MessageDb:
+
+`AccesStrategy.LatestKnownEvent`
+- Uses message-db's `get_last_stream_message` API to only ever fetch the last event in a stream
+- This is useful for aggregates whose entire state can be constructed from the latest event (e.g. a stream that stores checkpoints)
+- NOTE: The last event MUST be decodable by the supplied codec. Otherwise the stream is assumed to be empty
+
+`AccessStrategy.AdjacentSnapshots`
+- Generates and stores a snapshot event in an adjacent `{Category}:snapshot-{StreamId}` stream
+- The generation happens every `batchSize` events. This means the state of the stream can be reconstructed with exactly 2 round-trips to the database.
+  - The first round-trip fetches the most recent event of type `snapshotEventCaseName` from the snapshot stream.
+  - The second round-trip fetches `batchSize` events from the position of the snapshot
 
 <a name="dynamodb"></a>
 ### Use [Amazon DynamoDB](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Introduction.html)
--- a/README.md
+++ b/README.md
@@ -520,7 +520,7 @@ In addition to the default access strategy of reading the whole stream forwards 
 `AccesStrategy.LatestKnownEvent`
 - Uses message-db's `get_last_stream_message` API to only ever fetch the last event in a stream
 - This is useful for aggregates whose entire state can be constructed from the latest event (e.g. a stream that stores checkpoints)
-- NOTE: The last event MUST be decodable by the supplied codec. Otherwise the stream is assumed to be empty
+- NOTE: The last event should be decodable by the supplied codec. Otherwise you'll receive the initial state.
 
 `AccessStrategy.AdjacentSnapshots`
 - Generates and stores a snapshot event in an adjacent `{Category}:snapshot-{StreamId}` stream
--- a/samples/Store/Domain/Cart.fs
+++ b/samples/Store/Domain/Cart.fs
@@ -38,6 +38,7 @@ module Fold =
         let ofSnapshot (s: Events.Compaction.State) : State =
             { items = [ for i in s.items -> { skuId = i.skuId; quantity = i.quantity; returnsWaived = i.returnsWaived } ] }
     let initial = { items = [] }
+    // Should match the event type that is stored, which does not necessarily match the case name
     let snapshotEventCaseName = nameof Events.Snapshotted
     let evolve (state : State) event =
         let updateItems f = { state with items = f state.items }
--- a/samples/Store/Domain/Cart.fs
+++ b/samples/Store/Domain/Cart.fs
@@ -38,7 +38,7 @@ module Fold =
         let ofSnapshot (s: Events.Compaction.State) : State =
             { items = [ for i in s.items -> { skuId = i.skuId; quantity = i.quantity; returnsWaived = i.returnsWaived } ] }
     let initial = { items = [] }
-    let originEventType = nameof Events.Snapshotted
+    let snapshotEventCaseName = nameof Events.Snapshotted
     let evolve (state : State) event =
         let updateItems f = { state with items = f state.items }
         match event with
--- a/samples/Store/Domain/Cart.fs
+++ b/samples/Store/Domain/Cart.fs
@@ -38,7 +38,8 @@ module Fold =
         let ofSnapshot (s: Events.Compaction.State) : State =
             { items = [ for i in s.items -> { skuId = i.skuId; quantity = i.quantity; returnsWaived = i.returnsWaived } ] }
     let initial = { items = [] }
-    // Should match the event type that is stored, which does not necessarily match the case name
+    // NOTE Should match the event type that is stored, which does not necessarily match the case name
+    // e.g. if you override the name via [<DataMember(Name="snapshot-2")>], it needs to reflect that
     let snapshotEventCaseName = nameof Events.Snapshotted
     let evolve (state : State) event =
         let updateItems f = { state with items = f state.items }
--- a/samples/Store/Domain/Cart.fs
+++ b/samples/Store/Domain/Cart.fs
@@ -38,6 +38,7 @@ module Fold =
         let ofSnapshot (s: Events.Compaction.State) : State =
             { items = [ for i in s.items -> { skuId = i.skuId; quantity = i.quantity; returnsWaived = i.returnsWaived } ] }
     let initial = { items = [] }
+    let originEventType = nameof Events.Snapshotted
     let evolve (state : State) event =
         let updateItems f = { state with items = f state.items }
         match event with
--- a/src/Equinox.Core/Infrastructure.fs
+++ b/src/Equinox.Core/Infrastructure.fs
@@ -67,3 +67,4 @@ module Seq =
 module Array =
 
     let inline chooseV f xs = [| for item in xs do match f item with ValueSome v -> yield v | ValueNone -> () |]
+    let inline tryExactlyOneV (xs : _ array) = if xs.Length = 0 then ValueNone else ValueSome xs[0]
--- a/src/Equinox.Core/Infrastructure.fs
+++ b/src/Equinox.Core/Infrastructure.fs
@@ -67,4 +67,4 @@ module Seq =
 module Array =
 
     let inline chooseV f xs = [| for item in xs do match f item with ValueSome v -> yield v | ValueNone -> () |]
-    let inline tryExactlyOneV (xs : _ array) = if xs.Length = 0 then ValueNone else ValueSome xs[0]
+    let inline tryFirstV (xs : _ array) = if xs.Length = 0 then ValueNone else ValueSome xs[0]
--- a/src/Equinox.MessageDb/MessageDb.fs
+++ b/src/Equinox.MessageDb/MessageDb.fs
@@ -1,14 +1,14 @@
 ﻿namespace Equinox.MessageDb
 
-open System.Text.Json
 open Equinox.Core
+open Equinox.Core.Tracing
 open Equinox.MessageDb.Core
 open FsCodec
 open FsCodec.Core
 open Serilog
 open System
 open System.Diagnostics
-open Equinox.Core.Tracing
+open System.Text.Json
 
 type EventBody = ReadOnlyMemory<byte>
 
--- a/src/Equinox.MessageDb/MessageDb.fs
+++ b/src/Equinox.MessageDb/MessageDb.fs
@@ -126,7 +126,7 @@ module Log =
 
 module private Write =
 
-    let private writeEventsAsync (writer: MessageDbWriter) (streamName : string) (version : int64 option) (events : IEventData<EventBody> array)
+    let private writeEventsAsync (writer : MessageDbWriter) (streamName : string) (version : int64 option) (events : IEventData<EventBody> array)
         : Async<MdbSyncResult> = async {
         let! ct = Async.CancellationToken
         return! writer.WriteMessages(streamName, events, ct, ?version = version) |> Async.AwaitTaskCorrect }
@@ -181,7 +181,7 @@ module Read =
         let isLast = int64 page.Length < batchSize
         return toSlice page isLast }
 
-    let private readLastEventAsync (reader : MessageDbReader) (streamName : string) (requiresLeader : bool) (eventType: string option) ct = task {
+    let private readLastEventAsync (reader : MessageDbReader) (streamName : string) (requiresLeader : bool) (eventType : string option) ct = task {
         let! events = reader.ReadLastEvent(streamName, requiresLeader, ct, ?eventType = eventType)
         return toSlice events false }
 
@@ -297,7 +297,7 @@ module private Token =
 module private Snapshot =
     let inline snapshotCategory original = original + ":snapshot"
     let inline streamName category (streamId : string) = Equinox.Core.StreamName.render (snapshotCategory category) streamId
-    type Meta = {| streamVersion: int64 |} // STJ doesn't want to serialize it unless its anonymous
+    type Meta = {| streamVersion : int64 |} // STJ doesn't want to serialize it unless its anonymous
     let private streamVersion (evt: ITimelineEvent<EventBody>) =
         let meta = evt.Meta // avoid defensive copy
         JsonSerializer.Deserialize<Meta>(meta.Span).streamVersion
--- a/src/Equinox.MessageDb/MessageDb.fs
+++ b/src/Equinox.MessageDb/MessageDb.fs
@@ -384,6 +384,7 @@ type private Category<'event, 'state, 'context>(context : MessageDbContext, code
         | None -> context.LoadBatched(streamName, requireLeader, log, codec.TryDecode)
         | Some AccessStrategy.LatestKnownEvent -> context.LoadLast(streamName, requireLeader, log, codec.TryDecode)
         | Some (AccessStrategy.AdjacentSnapshots (snapshotType, _)) -> async {
+
             match! context.LoadSnapshot(category, streamId, requireLeader, log, codec.TryDecode, snapshotType) with
             | ValueSome (pos, snapshotEvent) ->
                 let! token, rest = context.Reload(streamName, requireLeader, log, pos, codec.TryDecode)
--- a/src/Equinox.MessageDb/MessageDb.fs
+++ b/src/Equinox.MessageDb/MessageDb.fs
@@ -420,8 +420,8 @@ type private Category<'event, 'state, 'context>(context : MessageDbContext, code
             match access with
             | None | Some AccessStrategy.LatestKnownEvent -> ()
             | Some (AccessStrategy.AdjacentSnapshots(_, toSnap)) ->
-              if Token.shouldSnapshot context.BatchOptions.BatchSize token token' then
-                do! x.StoreSnapshot(categoryName, streamId, log, ctx, token', toSnap state')
+                if Token.shouldSnapshot context.BatchOptions.BatchSize token token' then
+                    do! x.StoreSnapshot(categoryName, streamId, log, ctx, token', toSnap state')
             return SyncResult.Written   (token', state') }
 
 type private Folder<'event, 'state, 'context>(category : Category<'event, 'state, 'context>, fold : 'state -> 'event seq -> 'state, initial : 'state, ?readCache) =
--- a/src/Equinox.MessageDb/MessageDb.fs
+++ b/src/Equinox.MessageDb/MessageDb.fs
@@ -326,8 +326,7 @@ type MessageDbContext(connection : MessageDbConnection, batchOptions : BatchOpti
         let! _, events = Read.loadLastEvent log connection.ReadRetryPolicy connection.Reader requireLeader snapshotStream (Some eventType)
         match events with
         | [| event |] ->
-            let decoded = tryDecode event
-            match decoded with
+            match tryDecode event with
             | ValueSome decodedEvent ->
                 let meta = event.Meta
                 let meta = JsonSerializer.Deserialize<{| streamVersion: int64 |}>(meta.Span)
@@ -363,11 +362,16 @@ type AccessStrategy<'event, 'state> =
     /// doing a <c>fold</c> from any such event will yield a correct and complete state
     /// In other words, the <c>fold</c> function should not need to consider either the preceding <c>'state</c> or <c>'event</c>s.
     | LatestKnownEvent
-    /// Ensures a snapshot/compaction event from which the state can be reconstituted upon decoding is always present
-    /// (embedded in a special snapshot stream as an event), generated every <c>batchSize</c> events using the supplied
-    /// <c>toSnapshot</c> function.
-    /// When loading the stream it'll first fetch the snapshot.
-    | RollingSnapshots of originType: string * toSnapshot : ('state -> 'event)
+    /// <summary>
+    /// Generates and stores a snapshot event in an adjacent <c>{category}:snapshot-{stream_id}</c> stream
+    /// The generation happens every <c>batchSize</c> events.
+    /// This means the state of the stream can be reconstructed with exactly
+    /// 2 round-trips to the database.
+    /// The first round-trip fetches the most recent event of type <c>snapshotType</c> from the snapshot stream.
+    /// The second round-trip fetches <c>batchSize</c> events from the position of the snapshot
+    /// The snapshot is generated with the supplied <c>toSnapshot</c> function
+    /// </summary>
+    | AdjacentSnapshots of snapshotType: string * toSnapshot : ('state -> 'event)
 
 type private Category<'event, 'state, 'context>(context : MessageDbContext, codec : IEventCodec<_, _, 'context>, ?access: AccessStrategy<'event, 'state>) =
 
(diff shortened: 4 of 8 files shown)


tests/Equinox.EventStore.Integration/StoreIntegration.fs
```diff
@@ -339,12 +345,22 @@ type Tests(testOutputHelper) =
         let! _ = service.Read cartId
 
         // ... should see a single read as we are inside the batch threshold
+        #if STORE_MESSAGEDB
+        // The number of events is equal to the batch size, so it will append
+        // a snapshot event
+        test <@ readSnapshotted @ [EsAct.Append; EsAct.Append] @ readSnapshotted = capture.ExternalCalls @>
```
nordfjord, line 351: to appends. One for the stream, and one for the snapshot stream

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -281,6 +285,11 @@ module private Token =
           version = streamVersion + 1L
           streamBytes = -1 }
     let inline streamVersion (token : StreamToken) = unbox<int64> token.value
+    let shouldSnapshot batchSize prev next =
+        let previousVersion = prev.version
+        let nextVersion = next.version
+        let estimatedSnapshotPos = previousVersion - (previousVersion % batchSize)
+        nextVersion - estimatedSnapshotPos >= batchSize
```
nordfjord, line 292: naive implementation, but should suffice fine here

tests/Equinox.EventStore.Integration/StoreIntegration.fs
```diff
@@ -365,7 +367,53 @@ type Tests(testOutputHelper) =
         let! _ = service.Read cartId
         test <@ singleBatchBackwards @ batchBackwardsAndAppend @ singleBatchBackwards = capture.ExternalCalls @>
     }
-#endif
+    #else
+    [<AutoData(SkipIfRequestedViaEnvironmentVariable="EQUINOX_INTEGRATION_SKIP_EVENTSTORE")>]
+    let ``Can roundtrip against Store, correctly snapshotting to avoid redundant reads`` (ctx, skuId) = Async.RunSynchronously <| async {
```
nordfjord, line 372: I duplicated the compaction tests as it was too confusing with ifdefs.

It's different functionality and the times at which we snapshot will be different because there'll be a different number of events in the stream

bartelink: Makes sense

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -124,9 +126,11 @@ module Log =
 
 module private Write =
 
-    let private writeEventsAsync (writer: MessageDbWriter) (streamName : string) (version : int64) (events : IEventData<EventBody> array)
-        : Async<MdbSyncResult> =
-        writer.WriteMessages(streamName, events, version, Async.DefaultCancellationToken) |> Async.AwaitTaskCorrect
```
bartelink, line 129: waaait, how did I miss that!

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -311,8 +320,27 @@ type MessageDbContext(connection : MessageDbConnection, batchOptions : BatchOpti
         let! version, events = Read.loadForwardsFrom log connection.ReadRetryPolicy connection.Reader batchOptions.BatchSize batchOptions.MaxBatches streamName 0L requireLeader
         return Token.create version, Array.chooseV tryDecode events }
     member _.LoadLast(streamName, requireLeader, log, tryDecode) : Async<StreamToken * 'event array> = async {
-        let! version, events = Read.loadLastEvent log connection.ReadRetryPolicy connection.Reader requireLeader streamName
+        let! version, events = Read.loadLastEvent log connection.ReadRetryPolicy connection.Reader requireLeader streamName None
         return Token.create version, Array.chooseV tryDecode events }
+    member _.LoadSnapshot(category, streamId, requireLeader, log, tryDecode, eventType) = async {
+        let snapshotStream = StreamName.create (category + ":snapshot") streamId |> StreamName.toString
```
bartelink, line 326: For better or worse, equinox.core.streamname.render is used for this in most other places

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -311,8 +320,27 @@ type MessageDbContext(connection : MessageDbConnection, batchOptions : BatchOpti
...
+        let! _, events = Read.loadLastEvent log connection.ReadRetryPolicy connection.Reader requireLeader snapshotStream (Some eventType)
+        match events with
```
bartelink, line 328: events |> Array.tryExactyOne |> Option.bind tryDecode |> Option.map ... |> ValueOption.ofOption

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -311,8 +320,27 @@ type MessageDbContext(connection : MessageDbConnection, batchOptions : BatchOpti
...
+        | [| event |] ->
+            let decoded = tryDecode event
+            match decoded with
+            | ValueSome decodedEvent ->
+                let meta = event.Meta
+                let meta = JsonSerializer.Deserialize<{| streamVersion: int64 |}>(meta.Span)
```
bartelink, line 334: pull out a `module SnapshotStream` with `renderStreamName`, `parseVersion` etc

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -311,8 +320,27 @@ type MessageDbContext(connection : MessageDbConnection, batchOptions : BatchOpti
...
+                return ValueSome(Token.create meta.streamVersion, decodedEvent)
+            | ValueNone -> return ValueNone
+        | _ -> return ValueNone }
+
+    member _.StoreSnapshot(category, streamId, log, event) = async {
+        let snapshotStream = StreamName.create (category + ":snapshot") streamId |> StreamName.toString
+        do! Write.writeEvents log None connection.Writer snapshotStream None [| event |] |> Async.Ignore
```
bartelink, line 341: can remove `async {` wrapping

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -322,63 +350,94 @@ type MessageDbContext(connection : MessageDbConnection, batchOptions : BatchOpti
...
+type AccessStrategy<'event, 'state> =
     /// Load only the single most recent event defined in in a stream and trust that it'll be decoded and
     /// doing a <c>fold</c> from any such event will yield a correct and complete state
     /// In other words, the <c>fold</c> function should not need to consider either the preceding <c>'state</c> or <c>'event</c>s.
     | LatestKnownEvent
+    /// Ensures a snapshot/compaction event from which the state can be reconstituted upon decoding is always present
+    /// (embedded in a special snapshot stream as an event), generated every <c>batchSize</c> events using the supplied
+    /// <c>toSnapshot</c> function.
+    /// When loading the stream it'll first fetch the snapshot.
+    | RollingSnapshots of originType: string * toSnapshot : ('state -> 'event)
```
bartelink, line 370: needs a different name, maybe `SnapshotStream`

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -322,63 +350,94 @@ type MessageDbContext(connection : MessageDbConnection, batchOptions : BatchOpti
...
-type AccessStrategy =
+type AccessStrategy<'event, 'state> =
     /// Load only the single most recent event defined in in a stream and trust that it'll be decoded and
     /// doing a <c>fold</c> from any such event will yield a correct and complete state
     /// In other words, the <c>fold</c> function should not need to consider either the preceding <c>'state</c> or <c>'event</c>s.
     | LatestKnownEvent
+    /// Ensures a snapshot/compaction event from which the state can be reconstituted upon decoding is always present
+    /// (embedded in a special snapshot stream as an event), generated every <c>batchSize</c> events using the supplied
+    /// <c>toSnapshot</c> function.
+    /// When loading the stream it'll first fetch the snapshot.
```
bartelink, line 369:
```suggestion
    /// When loading the stream it'll first fetch attempt to fetch the first snapshot event of type <c>originType</c>.
```

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -322,63 +350,94 @@ type MessageDbContext(connection : MessageDbConnection, batchOptions : BatchOpti
...
             return GatewaySyncResult.Written token }
 
 [<NoComparison; NoEquality; RequireQualifiedAccess>]
-type AccessStrategy =
+type AccessStrategy<'event, 'state> =
     /// Load only the single most recent event defined in in a stream and trust that it'll be decoded and
     /// doing a <c>fold</c> from any such event will yield a correct and complete state
     /// In other words, the <c>fold</c> function should not need to consider either the preceding <c>'state</c> or <c>'event</c>s.
     | LatestKnownEvent
+    /// Ensures a snapshot/compaction event from which the state can be reconstituted upon decoding is always present
```
bartelink, line 366:
```suggestion
    /// Ensures availability of a snapshot from which the state can be efficiently reconstituted by only loading a single batch of events by generating and storing one in a sibling stream at intervals matching the <c>MessageDbContext</c>'s configured <c>batchSize</c>.
```

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -322,63 +350,94 @@ type MessageDbContext(connection : MessageDbConnection, batchOptions : BatchOpti
...
+    /// (embedded in a special snapshot stream as an event), generated every <c>batchSize</c> events using the supplied
```
bartelink, line 367: the above adjustment is triggered by use of the word embedded, which I was using to reflect that it's within the actual primary stream. Note also if the intellisense is to format the `<c>` pieces correctly, the entire comment needs to be wrapped in a `<summary>` tag (yes I have not done that in most places 'yet')

src/Equinox.MessageDb/MessageDb.fs
bartelink, line 370:
```suggestion
    | RollingSnapshots of snapshotCaseName : string * toSnapshot : ('state -> 'event)
```

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -322,63 +350,94 @@ type MessageDbContext(connection : MessageDbConnection, batchOptions : BatchOpti
...
 
-    let loadAlgorithm streamName requireLeader log =
+    let loadAlgorithm category streamId streamName requireLeader log =
         match access with
         | None -> context.LoadBatched(streamName, requireLeader, log, codec.TryDecode)
         | Some AccessStrategy.LatestKnownEvent -> context.LoadLast(streamName, requireLeader, log, codec.TryDecode)
+        | Some (AccessStrategy.RollingSnapshots (originType, _)) -> async {
+            let! latestSnapshotEvent = context.LoadSnapshot(category, streamId, requireLeader, log, codec.TryDecode, originType)
+            match latestSnapshotEvent with
+            | ValueSome (pos, ev) ->
```
bartelink, line 381:
```suggestion
            | ValueSome (pos, snapshotEvent) ->
```

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -322,63 +350,94 @@ type MessageDbContext(connection : MessageDbConnection, batchOptions : BatchOpti
...
-    member _.Load(fold : 'state -> 'event seq -> 'state, initial : 'state, stream, requireLeader, log : ILogger) =
-        load fold initial (loadAlgorithm stream requireLeader log)
+    member _.Load(fold : 'state -> 'event seq -> 'state, initial : 'state, category, streamId, streamName, requireLeader, log : ILogger) =
+        load fold initial (loadAlgorithm category streamId streamName requireLeader log)
     member _.Reload(fold : 'state -> 'event seq -> 'state, state : 'state, streamName, requireLeader, token, log : ILogger) =
         load fold state (context.Reload(streamName, requireLeader, log, token, codec.TryDecode))
 
+    member _.StoreSnapshot(category, streamId, log, ctx, token, snap) =
+        let encoded = codec.Encode(ctx, snap)
+        let meta = {| streamVersion = Token.streamVersion token; ``$correlationId`` = encoded.EventId; ``$causationId`` = encoded.EventId |}
```
bartelink, line 397: move into helper module with parsing

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -322,63 +350,94 @@ type MessageDbContext(connection : MessageDbConnection, batchOptions : BatchOpti
...
-                | ValueNone -> return! batched log streamName requireLeader ct
+                | ValueNone -> return! batched log categoryName streamId streamName requireLeader ct
                 | ValueSome tokenAndState when allowStale -> return tokenAndState
                 | ValueSome (token, state) -> return! category.Reload(fold, state, streamName, requireLeader, token, log) }
-        member _.TrySync(log, _categoryName, _streamId, streamName, context, _init, token, originState, events, _ct) = task {
-            match! category.TrySync(log, fold, streamName, token, originState, events, context) with
+
+        member _.TrySync(log, categoryName, streamId, streamName, context, _init, token, originState, events, ct) = task {
+            let! result = category.TrySync(log, fold, categoryName, streamId, streamName, token, originState, events, context) |> Async.startAsTask ct
+            match result with
```
bartelink, line 440: why you no like `match!` (there were even opportunities missed above!)

src/Equinox.MessageDb/Tracing.fs
```diff
@@ -19,7 +19,7 @@ type ActivityExtensions =
         act
 
     [<System.Runtime.CompilerServices.Extension>]
-    static member AddExpectedVersion(act: Activity, version: int64) =
+    static member AddExpectedVersion(act: Activity, version: int64 option) =
         act.AddTag("eqx.expected_version", version)
```
bartelink, line 23: will this render correctly with `null`, or should it become conditional?

tests/Equinox.EventStore.Integration/StoreIntegration.fs
```diff
@@ -365,7 +367,53 @@ type Tests(testOutputHelper) =
...
+        #if NET
+        use _ = source.StartActivity("Can roundtrip against Store, correctly snapshotting to avoid redundant reads")
+        #endif
+        let log, capture = output.CreateLoggerWithCapture()
+        let! client = connectToLocalStore log
+        let batchSize = 10
+        let context = createContext client batchSize
+        let service = Cart.createServiceWithCompaction log context
```
bartelink, line 380: differnt helper with different name so reader can identify its a different strategy at a glance
(maybe compaction term in this codebase should become embedded, and the other one termed siblingStream or something like that; conveying 'not embedded')

src/Equinox.MessageDb/MessageDb.fs
bartelink: note Rolling Snapshots, per the readme and/or DOC.md is referencing the semi-canonical definition of it in the esdb docs, which explicitly mentions embedding them in the main stream 

https://eventstore.org/docs/event-sourcing-basics/rolling-snapshots/index.html

All I could find was https://web.archive.org/web/20180626025148/https://eventstore.org/docs/event-sourcing-basics/rolling-snapshots/index.html

Perhaps this calls for the readme to have a section explaining snapshots can be:
- Embedded (SSS, ESDB) - needs to have a big warning about mixing state into event streams not being a bright idea in general
- Adjacent/Sibling - explains this is only available for mdb atm
- External (customisable external call - as that's not implemented, no need to mention)

Perhaps linking to this and using its terms might make sense ? https://www.eventstore.com/blog/snapshotting-strategies

Given we are going to V4 and Rolling Snapshots no longer seems to mean anything canonical anywhere, it might make sense to rename the DU cases for SSS and ESDB now to make it self-explanatory even without having read the relevant readme section

tests/Equinox.EventStore.Integration/StoreIntegration.fs
```diff
@@ -365,7 +367,53 @@ type Tests(testOutputHelper) =
...
+        capture.Clear()
+        let! _ = service.Read cartId
+        test <@ readSnapshotted = capture.ExternalCalls @>
+
+        // add two more events, triggering a snapshot, then read it in a single snapshotted read
+        capture.Clear()
+        do! addAndThenRemoveItemsManyTimes ctx cartId skuId service 1
+        // and reload the 20 events with a single read
+        let! _ = service.Read cartId
+        test <@ readSnapshotted @ [EsAct.Append; EsAct.Append] @ readSnapshotted = capture.ExternalCalls @>
```
bartelink, line 414: should there be assertions of how many events there are in the snapshots stream ?
I think not, but not 100% sure

tests/Equinox.EventStore.Integration/StoreIntegration.fs
```diff
@@ -457,7 +505,53 @@ type Tests(testOutputHelper) =
         test <@ [EsAct.AppendConflict; yield! sliceForward; EsAct.BatchForward] = capture.ExternalCalls @>
     }
 
-#if !STORE_MESSAGEDB
+#if STORE_MESSAGEDB
+    [<AutoData(SkipIfRequestedViaEnvironmentVariable="EQUINOX_INTEGRATION_SKIP_EVENTSTORE")>]
+    let ``Can combine snapshotting with caching against Store`` (ctx, skuId) = Async.RunSynchronously <| async {
```
bartelink, line 510:
```suggestion
    let ``Can combine snapshotting with caching against MessageDb Store`` (ctx, skuId) = Async.RunSynchronously <| async {
```
or maybe pull out a `type MessageDbTests` in this file, or a `MessageDbIntegration.fs`

bartelink:
Need to consider a naming strategy and/or using a custom category such that we don't pay for reading of snapshots in Propulsion.Mdb

(for cosmos, ddb, the reader excludes unfolds, although the cosmos changefeed will still charge you for reading its bodies).
For esdb, sss, there is no filtering atm given it's a rare thing)

(and maybe expose some predicate in a `.Internal` namespace if it turns out it needs to read them but skip them)

nordfjord:
> we don't pay for reading of snapshots in Propulsion.Mdb

I’m likely missing something, but you’d have to explicitly subscribe to `{Category}:snapshot` In propulsion to get snapshot events, so I’m not sure it’d be a problem.

bartelink:
> > we don't pay for reading of snapshots in Propulsion.Mdb
> 
> I’m likely missing something, but you’d have to explicitly subscribe to `{Category}:snapshot` In propulsion to get snapshot events, so I’m not sure it’d be a problem.

I probably mis-scanned the code - was assuming it was still in the same category as the events themselves - that's exactly what I meant in that case.

tests/Equinox.EventStore.Integration/StoreIntegration.fs
nordfjord: If I pull out a separate file should I go all the way and copy all 6 tests into it?

tests/Equinox.EventStore.Integration/StoreIntegration.fs
nordfjord: I'd be happy to leave that assertion out as ultimately the number of events doesn't matter, we only read the last one, as long as it is appended in one go (which is asserted on)

tests/Equinox.EventStore.Integration/StoreIntegration.fs
bartelink: probably not (I hadnt thought this point through in full when raising) - in that breakdown of the snapshotting mechanisms, you have 
a) adjacent ones (mdb only for now, but could add to esdb)
b) embedded (esdb, sss)
c) unfolds based

src/Equinox.MessageDb/Tracing.fs
nordfjord: Screenshot 2022-11-21 at 9 06 45 AM
Thanks for shouting. It correctly excludes `None` but renders the some explicitly as `Some(version)` 😬

tests/Equinox.EventStore.Integration/StoreIntegration.fs
bartelink: sounds about right, I guess the best thing to predicate things on is whether the correct number of slice reads happen (exactly one) based on the intended timing of when snapshots get written.

src/Equinox.MessageDb/MessageDb.fs
nordfjord: Got tired of explaining "It's just a let! binding followed by a match expression" 😂.

I'll change

src/Equinox.MessageDb/MessageDb.fs
nordfjord: how did I write that!? 😄

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -363,21 +362,25 @@ type AccessStrategy<'event, 'state> =
...
+    /// <summary>
+    /// Generates and stores a snapshot event in an adjacent <c>{category}:snapshot-{stream_id}</c> stream
+    /// The generation happens every <c>batchSize</c> events.
+    /// This means the state of the stream can be reconstructed with exactly
+    /// 2 round-trips to the database.
+    /// The first round-trip fetches the most recent event of type <c>snapshotType</c> from the snapshot stream.
+    /// The second round-trip fetches <c>batchSize</c> events from the position of the snapshot
+    /// The snapshot is generated with the supplied <c>toSnapshot</c> function
+    /// </summary>
+    | AdjacentSnapshots of snapshotType: string * toSnapshot : ('state -> 'event)
```
bartelink, line 374: snapshotType still needs a better name that conveys its the caseName associated with a snapshot DU case (ideally the same one as `toSnapshot`).

🤔 I wonder if could it be inferred from an Expr<'event> that creates a dummy value e.g. `<@ fun () -> Snapshot Unchehecked.defaultof<_> @>`

samples/Store/Domain/Cart.fs
```diff
@@ -38,6 +38,7 @@ module Fold =
         let ofSnapshot (s: Events.Compaction.State) : State =
             { items = [ for i in s.items -> { skuId = i.skuId; quantity = i.quantity; returnsWaived = i.returnsWaived } ] }
     let initial = { items = [] }
+    let originEventType = nameof Events.Snapshotted
```
bartelink, line 41: snapshotEventCaseName ?

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -311,74 +319,127 @@ type MessageDbContext(connection : MessageDbConnection, batchOptions : BatchOpti
...
+type AccessStrategy<'event, 'state> =
     /// Load only the single most recent event defined in in a stream and trust that it'll be decoded and
     /// doing a <c>fold</c> from any such event will yield a correct and complete state
     /// In other words, the <c>fold</c> function should not need to consider either the preceding <c>'state</c> or <c>'event</c>s.
     | LatestKnownEvent
-
-type private Category<'event, 'state, 'context>(context : MessageDbContext, codec : IEventCodec<_, _, 'context>, ?access) =
-
-    let loadAlgorithm streamName requireLeader log =
+    /// <summary>
```
bartelink, line 365: This reads very well now 👍

src/Equinox.MessageDb/MessageDb.fs
nordfjord: A bit of Maths™️  😅

We want to snapshot every `B`atchsize events, and version is function of the number of events in the stream

Given a version I can therefore know when the last snapshot occurred with a bit of modulo magic

`version - (version % batchsSize)`

e.g. if the batchsize is 10, and the version is 15 this yield `15 - (15 % 10) = 15 - 5 = 10`

 Now that we have the previous snapshot placement we know if we have to snapshot based on the difference between the snapshot version and the current version.

src/Equinox.MessageDb/MessageDb.fs
nordfjord: > 🤔 I wonder if could it be inferred from an Expr<'event> that creates a dummy value e.g. <@ fun () -> Snapshot Unchehecked.defaultof<_> @>

That might be a bit much to ask of a consumer of the library

> snapshotType still needs a better name that conveys its the caseName associated with a snapshot DU case (ideally the same one as toSnapshot).

I'll ask around what people think makes sense, for now I've renamed to "snapshotEventCaseName"

bartelink: >That might be a bit much to ask of a consumer of the library

yes, was trying to provoke thought by putting something pejorative out, but there has to be some static reflection type trick of some kind!
I personally would use a `[<>]` literal in the `module Events` and use to it in Fold.snapshoteventcasename if it arose, but then I'm not the person that will whack it in and not understand its role, either

src/Equinox.MessageDb/MessageDb.fs
bartelink: >A bit of Maths™️ 😅

and there's absolutely definitely no way you could be off by one, ever?
and the previous version is definitely from the snapshot, and not from the last cached append (it scanned fine, but I'd be pressing * and # for a long time making sure!)

samples/Store/Domain/Cart.fs
bartelink: maybe put a comment in there explaining that it might not always be `nameof` ?

src/Equinox.MessageDb/MessageDb.fs
nordfjord: > and there's absolutely definitely no way you could be off by one, ever?

the formula is sound. But I cannot guarantee absence of off-by-ones. I could be off by one with a `>` vs `>=`, but I think that would've been caught in the integration test (with an additional slice read being necessary)

bartelink: Yes, not doubting you or the formula, which seems correct and is obviously not the most mind bending math EVAR. Just asking the rhetorical question about whether such edge cases are covered. In general the cases you took from try to exercise snapshots being 'just' due or 'just not yet', but e.g. for ESDB, the token tests cover a lot of these cases.

samples/Store/Domain/Cart.fs
nordfjord: Is there something we can use from TypeShape or FsCodec? `caseNameOf Event`?

How does FsCodec decide what string to use for the event type? Can we hook into that?

nordfjord: Alternatively, this might not be the worst idea

```fsharp
let snapshotEventCaseName = codec.Encode((), Snapshotted).EventType 
```

nordfjord: Oh this falls into a "fun with reflection" territory I'm not super comfortable with 😅 

https://github.com/eiriktsarpalis/TypeShape/blob/main/src/TypeShape/Applications/UnionContract.fs#L97-L100

I think we could extract something like this:

```fsharp
let getCaseName case =
        let t = case.GetType()
        let cases = FSharpType.GetUnionCases(t, true)
        let tagIdx = FSharpValue.PreComputeUnionTagReader(t, true) case
        let caseInfo = cases[tagIdx]
        caseInfo.GetCustomAttributes()
        |> Seq.tryPick (function :? DataMemberAttribute as dm -> Some dm.Name | _ -> None)
        |> Option.defaultValue(caseInfo.Name)
```

bartelink: yes, ideally that'd go as a PR upstream

Also in the [docs wouldnt it be nice section](https://github.com/jet/equinox/blob/master/DOCUMENTATION.md#wouldnt-it-be-nice---equinoxeventstoredb) I have:

> Provide a facility in FsCodec.IEventCodec to walk the Event DU to generate a list of event types; use that to generate the server-side event loading filter e.g. when a Decider used a highly selective subset of the known Event Types

i.e. if it can be generalised in some way, that'd be nice

bartelink: (none of this should be a blocker; it's perfectly reasonable to file it as an issue here and/or in TypeShape)

nordfjord: > i.e. if it can be generalised in some way, that'd be nice

I think it can be.

```fsharp
let caseName (caseInfo: UnionCaseInfo) =
    caseInfo.GetCustomAttributes()
    |> Seq.tryPick (function :? DataMemberAttribute as dm -> Some dm.Name | _ -> None)
    |> Option.defaultValue(caseInfo.Name)
let getCaseNames<'t> =
    let t = typeof<'t>
    let cases = FSharpType.GetUnionCases(t, true)
    [| for caseInfo in cases -> caseName caseInfo |]
```

tests/Equinox.EventStore.Integration/StoreIntegration.fs
nordfjord: Seeing as we have RollingSnapshots being tested in this file already, I think we might want to keep these all in the same file. I'll pull it out into a separate `type`

nordfjord: I've pulled out `GeneralTests` `RollingSnapshotsTest` and `AdjacentSnapshotTest` into their own types

tests/Equinox.EventStore.Integration/StoreIntegration.fs
```diff
@@ -503,27 +557,118 @@ type Tests(testOutputHelper) =
...
+           .AddOtlpExporter(fun opts -> opts.Endpoint <- Uri("http://localhost:4317"))
+           .Build()
+
+    let output = TestContext(testOutputHelper)
+
+    let sliceForward = [EsAct.SliceForward]
+    let singleBatchForward = [EsAct.SliceForward; EsAct.BatchForward]
+    let readSnapshotted = [EsAct.ReadLast; EsAct.SliceForward; EsAct.BatchForward]
+
     [<AutoData(SkipIfRequestedViaEnvironmentVariable="EQUINOX_INTEGRATION_SKIP_EVENTSTORE")>]
```
bartelink, line 580: maybe this can now be to be `EQUINOX_INTEGRATION_SKIP_MESSAGEDB`, with associated def of that in the CI scripts

tests/Equinox.EventStore.Integration/StoreIntegration.fs
bartelink: looks good

tests/Equinox.EventStore.Integration/StoreIntegration.fs
nordfjord: It probably can be. Though since the purpose is to exclude from CI, I think we might as well change the variable to "CI" as that's set automatically in azure and github actions.

nordfjord: Suggestion: Let's keep as is, and my next PR will rig CI for message-db and SSS PG?

nordfjord: Just conscious of not doing too much in the span of a single PR

bartelink: no arguments of any kind :D Also no pressure...

samples/Store/Domain/Cart.fs
```diff
@@ -38,6 +38,8 @@ module Fold =
         let ofSnapshot (s: Events.Compaction.State) : State =
             { items = [ for i in s.items -> { skuId = i.skuId; quantity = i.quantity; returnsWaived = i.returnsWaived } ] }
     let initial = { items = [] }
+    // Should match the event type that is stored, which does not necessarily match the case name
```
bartelink, line 41:
```suggestion
    // NOTE Should match the event type that is stored, which does not necessarily match the case name
    // e.g. if you override the name via [<DataMember(Name="snapshot-2")>], it needs to reflect that
```

src/Equinox.Core/Infrastructure.fs
```diff
@@ -67,3 +67,4 @@ module Seq =
 module Array =
 
     let inline chooseV f xs = [| for item in xs do match f item with ValueSome v -> yield v | ValueNone -> () |]
+    let inline tryExactlyOneV (xs : _ array) = if xs.Length = 0 then ValueNone else ValueSome xs[0]
```
bartelink, line 70: (otherwise this should be tryFirstV)
```suggestion
    let inline tryExactlyOneV (xs : _ array) = if xs.Length <> 1 then ValueNone else ValueSome xs[0]
```

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -1,8 +1,10 @@
 ﻿namespace Equinox.MessageDb
 
+open System.Text.Json
```
bartelink, line 3: sort

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -124,9 +126,11 @@ module Log =
 
 module private Write =
 
-    let private writeEventsAsync (writer: MessageDbWriter) (streamName : string) (version : int64) (events : IEventData<EventBody> array)
-        : Async<MdbSyncResult> =
-        writer.WriteMessages(streamName, events, version, Async.DefaultCancellationToken) |> Async.AwaitTaskCorrect
+    let private writeEventsAsync (writer: MessageDbWriter) (streamName : string) (version : int64 option) (events : IEventData<EventBody> array)
```
bartelink, line 129:
```suggestion
    let private writeEventsAsync (writer : MessageDbWriter) (streamName : string) (version : int64 option) (events : IEventData<EventBody> array)
```

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -178,8 +181,8 @@ module Read =
         let isLast = int64 page.Length < batchSize
         return toSlice page isLast }
 
-    let private readLastEventAsync (reader : MessageDbReader) (streamName : string) (requiresLeader : bool) ct = task {
-        let! events = reader.ReadLastEvent(streamName, requiresLeader, ct)
+    let private readLastEventAsync (reader : MessageDbReader) (streamName : string) (requiresLeader : bool) (eventType: string option) ct = task {
```
bartelink, line 184:
```suggestion
    let private readLastEventAsync (reader : MessageDbReader) (streamName : string) (requiresLeader : bool) (eventType : string option) ct = task {
```

src/Equinox.MessageDb/MessageDb.fs
bartelink, line 129: make this be `task{` like the equivalent level in the ReadLastEvent chain (triggered by Async in name of an Async fn - I tend to only ever use that for Task stuff these days)

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -281,10 +284,28 @@ module private Token =
...
+        nextVersion - estimatedSnapshotPos >= batchSize
 
     let supersedes struct (current, x) =
         x.version > current.version
 
+
+module private Snapshot =
+    let inline snapshotCategory original = original + ":snapshot"
+    let inline streamName category (streamId : string) = Equinox.Core.StreamName.render (snapshotCategory category) streamId
+    type Meta = {| streamVersion: int64 |} // STJ doesn't want to serialize it unless its anonymous
```
bartelink, line 300:
```suggestion
    type Meta = {| streamVersion : int64 |} // STJ doesn't want to serialize it unless its anonymous
```

src/Equinox.Core/Infrastructure.fs
nordfjord: 🤔 We should only ever get one event, but I'm not sure if we want to short circuit and ignore if we ever get more than one. Might as well just take the first in that case. I'll make it `tryFirstV`

src/Equinox.MessageDb/MessageDb.fs
nordfjord: Not sure this function even needs to exist 🤔 It's just calling the method on the writer

nordfjord: I might do a cleanup later, will change this to task for now

src/Equinox.MessageDb/MessageDbClient.fs
```diff
@@ -32,29 +42,31 @@ module private Npgsql =
...
         cmd.Parameters.AddWithValue("EventType", NpgsqlDbType.Text, e.EventType) |> ignore
         cmd.Parameters |> Json.addParameter "Data" e.Data
         cmd.Parameters |> Json.addParameter "Meta" e.Meta
-        cmd.Parameters.AddWithValue("ExpectedVersion", NpgsqlDbType.Bigint, expectedVersion) |> ignore
+        cmd.Parameters |> Sql.addNullableInt64 "ExpectedVersion" expectedVersion
 
         cmd
 
-    member _.WriteMessages(streamName, events : _ array, version : int64, ct) = task {
+    member _.WriteMessages(streamName, events : _ array, ct, ?version : int64) = task {
```
nordfjord, line 57: I hate this api, version shouldn't be an optional arg, it's an option

src/Equinox.MessageDb/MessageDb.fs
bartelink: yes this layer is only really sensible if there is no Client construxt

src/Equinox.MessageDb/MessageDbClient.fs
bartelink: yeah, should ahve trapped earlier

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -311,74 +332,118 @@ type MessageDbContext(connection : MessageDbConnection, batchOptions : BatchOpti
...
-    member _.Load(fold : 'state -> 'event seq -> 'state, initial : 'state, stream, requireLeader, log : ILogger) =
-        load fold initial (loadAlgorithm stream requireLeader log)
+    member _.Load(fold : 'state -> 'event seq -> 'state, initial : 'state, category, streamId, streamName, requireLeader, log : ILogger) =
+        load fold initial (loadAlgorithm category streamId streamName requireLeader log)
     member _.Reload(fold : 'state -> 'event seq -> 'state, state : 'state, streamName, requireLeader, token, log : ILogger) =
         load fold state (context.Reload(streamName, requireLeader, log, token, codec.TryDecode))
 
+    member _.StoreSnapshot(category, streamId, log, ctx, token, snap) =
+        let encoded = codec.Encode(ctx, snap)
+        let meta = {| streamVersion = Token.streamVersion token |}
```
bartelink, line 402: this but prob also 406 should go into module Snapshot

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -311,74 +332,118 @@ type MessageDbContext(connection : MessageDbConnection, batchOptions : BatchOpti
...
+    /// </summary>
+    | AdjacentSnapshots of snapshotEventCaseName: string * toSnapshot : ('state -> 'event)
+
+type private Category<'event, 'state, 'context>(context : MessageDbContext, codec : IEventCodec<_, _, 'context>, ?access: AccessStrategy<'event, 'state>) =
+
+    let loadAlgorithm category streamId streamName requireLeader log =
         match access with
         | None -> context.LoadBatched(streamName, requireLeader, log, codec.TryDecode)
         | Some AccessStrategy.LatestKnownEvent -> context.LoadLast(streamName, requireLeader, log, codec.TryDecode)
+        | Some (AccessStrategy.AdjacentSnapshots (snapshotType, _)) -> async {
```
bartelink, line 384: this all reads very clean, nice

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -311,74 +330,118 @@ type MessageDbContext(connection : MessageDbConnection, batchOptions : BatchOpti
...
         | GatewaySyncResult.ConflictUnknown ->
             return SyncResult.Conflict  (fun ct -> x.Reload(fold, state, streamName, (*requireLeader*)true, token, log) |> Async.startAsTask ct)
         | GatewaySyncResult.Written token' ->
-            return SyncResult.Written   (token', fold state (Seq.ofArray events)) }
+            let state' = fold state (Seq.ofArray events)
+            match access with
+            | None | Some AccessStrategy.LatestKnownEvent -> ()
+            | Some (AccessStrategy.AdjacentSnapshots(_, toSnap)) ->
+              if Token.shouldSnapshot context.BatchOptions.BatchSize token token' then
+                do! x.StoreSnapshot(categoryName, streamId, log, ctx, token', toSnap state')
```
bartelink, line 424:
```suggestion
                  do! x.StoreSnapshot(categoryName, streamId, log, ctx, token', toSnap state')
```
need editor config!

nordfjord: The if as well?

bartelink: heck, lets go all out, yep!

README.md
```diff
@@ -513,7 +513,20 @@ MessageDb support is provided in the samples and the `eqx` tool:
...
 
-Equinox does not provide utilities for configuring or installing message-db. See [MessageDB's installation documentation](http://docs.eventide-project.org/user-guide/message-db/install.html).
+Equinox does not provide utilities for configuring or installing MessageDB. See [MessageDB's installation documentation](http://docs.eventide-project.org/user-guide/message-db/install.html).
+
+In addition to the default access strategy of reading the whole stream forwards in batches, the following access strategies are supported in MessageDb:
+
+`AccesStrategy.LatestKnownEvent`
+- Uses message-db's `get_last_stream_message` API to only ever fetch the last event in a stream
+- This is useful for aggregates whose entire state can be constructed from the latest event (e.g. a stream that stores checkpoints)
+- NOTE: The last event MUST be decodable by the supplied codec. Otherwise the stream is assumed to be empty
```
bartelink, line 523: should be mentioned in the xmldoc if not yet

README.md
```diff
@@ -513,7 +513,20 @@ MessageDb support is provided in the samples and the `eqx` tool:
 - being able to supply `mdb` flag to `eqx dump`, e.g. `eqx dump -CU "Favorites-ab25cc9f24464d39939000aeb37ea11a" mdb -c "pgconnectionstring"`
 - being able to supply `mdb` flag to Web sample, e.g. `dotnet run --project samples/Web/ -- mdb -c "pgconnectionstring"`
 
-Equinox does not provide utilities for configuring or installing message-db. See [MessageDB's installation documentation](http://docs.eventide-project.org/user-guide/message-db/install.html).
+Equinox does not provide utilities for configuring or installing MessageDB. See [MessageDB's installation documentation](http://docs.eventide-project.org/user-guide/message-db/install.html).
+
+In addition to the default access strategy of reading the whole stream forwards in batches, the following access strategies are supported in MessageDb:
```
bartelink, line 518: will probably mention this above https://github.com/jet/equinox/blob/c9d689926e1a1a147febf8c3d9de2286a92e3696/README.md#currently-supported-data-stores together with a mention of it having otel wiring

README.md
nordfjord: Clarified that it's not empty, but that it's only a single event, and therefore you'll get the initial state if the last event isn't decodable. MUST -> should

PR #357, opened 2022-11-21
Esdb, Sss, Mdb: Implement ITimelineEvent.Size

bartelink:
Adds a baseline event size computation for the specified stores
Required for correct StreamsSink state logging when feeding from `EventStoreDbSource` https://github.com/jet/propulsion/pull/185

### DIFF
--- a/src/Equinox.MessageDb/MessageDbClient.fs
+++ b/src/Equinox.MessageDb/MessageDbClient.fs
@@ -1,7 +1,6 @@
 namespace Equinox.MessageDb.Core
 
 open FsCodec
-open FsCodec.Core
 open Npgsql
 open NpgsqlTypes
 open System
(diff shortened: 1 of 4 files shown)


src/Equinox.MessageDb/MessageDbClient.fs
```diff
@@ -64,16 +65,13 @@ type MessageDbReader internal (connectionString : string, leaderConnectionString
 
     let parseRow (reader : DbDataReader) : ITimelineEvent<Format> =
         let inline readNullableString idx = if reader.IsDBNull(idx) then None else Some (reader.GetString idx)
-
-        TimelineEvent.Create(
+        let c, d, m = reader.GetString(1), reader |> Json.fromReader 2, reader |> Json.fromReader 3
```
nordfjord, line 68: Not sure this is an improvement readability wise 🤔 can we keep it as `eventType`, `data`, and `meta`?

nordfjord: Not a blocker at all, can see the argument for keeping it unified across the stores as well

bartelink: that was what I was driving at, and I do realise that we can afford the space
the c for case is a bit ugly/debatable, but it prepares people for the other stores
lets see how I feel after my run!

bartelink: Problem solved, and I saw a stupid bug I had introduced

PR #364, opened 2022-12-12
Upgrade Npgsql to v7 now that it's GA

nordfjord:
Npgsql has released v7, so let's get up to date in MessageDB


bartelink:
Thanks!
NOTE (for other readers more than for you) in general we don't tend to update dependencies in Equinox libs from the point of view of trying to reflect the recommended minimum client library version, except for generally doing a bump when there's a major version release.
The assumption implied is that consumer apps should in general be proactively tracking latest versions possible across their full set of dependencies, and upgrading to reflect that on a continual basis independent of Equinox releases, etc.
In this case it makes sense to up it as it's a fresh component that's yet to have it's first non `-rc` release.

nordfjord:
I'll keep that in mind for future reference :+1:

PR #365, opened 2022-12-21
MessageDb: Remove Connector

bartelink:
Aligns MessageDb with DynamoStore and CosmosStore in using `Client` to refer to the layer that talks to the underlying store (but does not hold active connections as such)
- Having a Client similarly implies "make one of these and reuse them"
- However having an Establish or Connect method suggests that there is a long lived connection element and/or something interesting going on (for EventStore, Establish was originally a one-off Async method that did something)


bartelink:
(Ran into this in the course of revisiting a name for DynamoStore - open to leaving it as it is if anyone thinks it was better the way it was)

PR #367, opened 2023-01-30
DynamoDb: Fix integration test expectations

bartelink:
The repo contains a docker compose file that provisions a dynamodb-local instance
For the last batch of edits, I tested against a DynamoDB cloud instance

I erroneously updated the test expectations to match what I observed without considering that:
- I had tested against a cloud instance (the expectation is that the tests should just work from trunk).
- _dynamodb-local continues to have this behavioral difference_ (if you query with a maxItems that coincides with the number of items in the partition, you still get yielded a continuation token, which triggers an extra roundtrip)

This change makes the expectations of the integration tests suite conditional on the target instance URL, which is hacky. Hopefully this simulator difference goes away eventually...

resolves #366 (🙏 @nordfjord who did all the research to determine the problem, and when it was introduced)

### DIFF
--- a/tests/Equinox.CosmosStore.Integration/CosmosFixtures.fs
+++ b/tests/Equinox.CosmosStore.Integration/CosmosFixtures.fs
@@ -15,12 +15,13 @@ let discoverConnection () =
     match tryRead "EQUINOX_DYNAMO_SERVICE_URL" with // NOT USING EQUINOX_DYNAMO_SERVICE_URL env var as we don't want to go provisioning 2 tables in a random DB
     | None -> "dynamodb-local", "http://localhost:8000"
     | Some connectionString -> "EQUINOX_DYNAMO_CONNECTION", connectionString
+let isSimulatorServiceUrl url = Uri(url).IsLoopback
 
 let createClient (log : Serilog.ILogger) name serviceUrl =
     // See https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.DownloadingAndRunning.html#docker for details of how to deploy a simulator instance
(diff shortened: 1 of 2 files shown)


tests/Equinox.CosmosStore.Integration/CosmosFixtures.fs
```diff
@@ -12,18 +12,21 @@ let private tableName = tryRead "EQUINOX_DYNAMO_TABLE" |> Option.defaultValue "e
...
 
 let createClient (log : Serilog.ILogger) name serviceUrl =
     // See https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBLocal.DownloadingAndRunning.html#docker for details of how to deploy a simulator instance
     let clientConfig = AmazonDynamoDBConfig(ServiceURL = serviceUrl)
     log.Information("DynamoStore {name} {endpoint}", name, serviceUrl)
-    // Credentials are not validated if connecting to local instance so anything will do (this avoids it looking for profiles to be configured)
-    // OR: don't pass credentials to ctor to use keychain configured access
-    let credentials = Amazon.Runtime.BasicAWSCredentials("A", "A")
-    new AmazonDynamoDBClient(credentials, clientConfig) :> IAmazonDynamoDB
+    if serviceUrl.Contains "localhost" then
```
nordfjord, line 23:
```suggestion
    if System.Uri(serviceUrl).IsLoopback then
```

Just in case `::1` or `127.0.01` 😄

nordfjord: ^ also, don't. This doesn't matter. Just felt like being a pedant

PR #371, opened 2023-03-14
fix issue with uncaught StreamNotFoundException after migration to Ta…

### LINKED ISSUE #370: StreamNotFoundException when trying to add Events to a new Stream with EventStoreDb
When trying to use Equinox EventStoreDb I encountered a StreamNotFoundException whenever I tried to add the first Event to a new Stream. (Strangely creating the first Event and thus creating the Stream worked via the AdminUI. Also appending to this Stream with my implementation worked fine after creating the first event via the AdminUI).

Since I'm pretty much new to Equinox my first guess was that I configured something wrong. So I tried to setup the "Backend Todo" dotnet template and figure it out myself. But here I encountered the exact same error.

Steps to reproduce:
1. Follow instructions of this: https://github.com/jet/equinox#store-data-in-eventstore
(Note that I did use 22.10.1-buster-slim instead of 21.10.0-buster-slim, but I figured this should actually be the correct version since Equinox.EventStore is using the Client for Version 22.0.0-preview https://github.com/jet/equinox/blob/master/src/Equinox.EventStore/Equinox.EventStore.fsproj#L23 )
2. Run a POST HTTP request against http://localhost:5000/todos with the following payload
`{
    "id": 0,
    "url": "string",
    "order": 0,
    "title": "string",
    "completed": false
}`

When I had a quick look at the code line where the Exception was thrown I saw that the wrong Exception got caught here.

Code line in EventStoreDb.fs where the Exception was thrown:
https://github.com/jet/equinox/blob/d5cc2c92028405900daeaaf5bac2bca7f077ebe3/src/Equinox.EventStoreDb/EventStoreDb.fs#L198

When I replaced this line with
`with :? StreamNotFoundException)`
it worked fine in the "Backend Todo" example.

Is it possible that the Exception signature here is simply no up to date with the latest EventStore version?
See: https://github.com/EventStore/EventStore-Client-Dotnet/blob/master/src/EventStore.Client.Streams/EventStoreClient.Read.cs#L353

Hope this is somewhat helpful information to you and not just a mistake on my side. If my assumption is correct then I am happy to create a Pull-Request :)

keppelerj:
…skSeq
Closes #370

### DIFF
--- a/src/Equinox.EventStoreDb/EventStoreDb.fs
+++ b/src/Equinox.EventStoreDb/EventStoreDb.fs
@@ -168,6 +168,13 @@ module private Write =
 
 module private Read =
     open FSharp.Control
+    let private processReadStreamResult (res: EventStoreClient.ReadStreamResult) f onStreamNotFound = task {
+        let! readState = res.ReadState
+        match readState with
+        | ReadState.StreamNotFound -> return onStreamNotFound
+        | ReadState.Ok -> return! res |> f
+    }
+
     let resolvedEventBytes (x : ResolvedEvent) = let Log.BlobLen bytes, Log.BlobLen metaBytes = x.Event.Data, x.Event.Metadata in bytes + metaBytes
     let resolvedEventsBytes events = events |> Array.sumBy resolvedEventBytes
     let private logBatchRead direction streamName t events batchSize version (log : ILogger) =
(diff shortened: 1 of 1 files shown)


src/Equinox.EventStoreDb/EventStoreDb.fs
```diff
@@ -181,36 +188,43 @@ module private Read =
...
-                        false
-                    | _ -> true)
-                |> TaskSeq.toArrayAsync
-            let v = match Seq.tryHead events with Some (r, _) -> let en = r.Event.EventNumber in en.ToInt64() | None -> -1
-            Array.Reverse events
-            return v, events
-        with :? AggregateException as e when (e.InnerExceptions.Count = 1 && e.InnerExceptions[0] :? StreamNotFoundException) ->
-            return -1L, [||] }
+        let processResult (result: EventStoreClient.ReadStreamResult) =
+            task {
```
bartelink, line 194: for better or for worse, `task {` is always at the end of the preceding line in this repo

src/Equinox.EventStoreDb/EventStoreDb.fs
```diff
@@ -168,6 +168,13 @@ module private Write =
 
 module private Read =
     open FSharp.Control
+    let private processReadStreamResult (res: EventStoreClient.ReadStreamResult) f onStreamNotFound = task {
+        let! readState = res.ReadState
+        match readState with
```
bartelink, line 173: collapse as `match!`

src/Equinox.EventStoreDb/EventStoreDb.fs
```diff
@@ -168,6 +168,13 @@ module private Write =
 
 module private Read =
     open FSharp.Control
+    let private processReadStreamResult (res: EventStoreClient.ReadStreamResult) f onStreamNotFound = task {
```
bartelink, line 171: Scanning this right now in web view, I'm not convinced having a separated function makes the overall code comprehension exercise easier - i.e. inlining instead of being DRY might be clearer? (Will look in an IDE later)

src/Equinox.EventStoreDb/EventStoreDb.fs
keppelerj: oh nice didn't know about that :)

src/Equinox.EventStoreDb/EventStoreDb.fs
keppelerj: Yes maybe you're right about. I'm also not happy with the function name tbh

bartelink: I inlined it in https://github.com/jet/equinox/pull/373/commits/740cc4a861f7f851641a7bb05d4e9234877229f1
If you can cherry pick that and/or do similar in your own context, I'll release (or say the word and I can merge yours and apply it as a fixup on master afterwards)

src/Equinox.EventStoreDb/EventStoreDb.fs
```diff
@@ -168,6 +168,12 @@ module private Write =
...
+        match! res.ReadState with
+        | ReadState.StreamNotFound -> return onStreamNotFound
+        | ReadState.Ok -> return! f res
```
bartelink, line 174: note for .NET Enum values, matching like this produces a warning to indicate that there is a potential for an out of range enum value. While I generally would not do a catchall, I think that makes the most sense for a case like this, and it removes the warning.
(You could also layer an Active Pattern over the enum to codify that - will try that now)

src/Equinox.EventStoreDb/EventStoreDb.fs
keppelerj: Thank you! Cherry picked it into this branch

src/Equinox.EventStoreDb/EventStoreDb.fs
keppelerj: Dang, these evil .NET enums.. 
When layering an active pattern over the enum wouldn't you encounter the same issue just at a different position? (At the declaration of active pattern). But it might still be better and more reusable.

bartelink: Main issue with that active pattern is that it allocates, and would be crazy to do in a tight loop.
As you say, you still have a catchall that would need to be revisited if they ever added a (meaningful) new enum value.
Given there are only 2 usages, it would definitely be overkill here - the only value would be that you get to encapsulate and DRY the mapping of relevant cases.

bartelink:
Thanks, will merge/release in a few hours...

keppelerj:
Thank you very much! No hurries :)

src/Equinox.EventStoreDb/EventStoreDb.fs
keppelerj: Ah ok I see! Didn't know about the performance downsides of active patterns (i'm still pretty new to the f# world). Thanks again for the insights!

bartelink:
Release `4.0.0-rc8.1` is on nuget now, thanks again for the work

PR #374, opened 2023-03-31
Cosmos: Add Docker emulator config

bartelink:
Adds `docker compose` config for the CosmosDB emulator [for a Mac]
- [ ] 2 failing tests (re Lazy semantics) awaiting updated `FSharp.Control.TaskSeq` that reinstates `try/finally` when a sequence is not fully drained


bartelink:
NOTE at present, there are 2 failing tests for Equinox.Cosmos (re Lazy semantics) awaiting an updated `FSharp.Control.TaskSeq` release that runs `try/finally` correctly (both when fully drained, and when a sequence is not fully drained)

PR #375, opened 2023-04-04
Add Transact*Task api's to the F# decider abstraction

nordfjord:
We're moving towards using Tasks more and more and this API would be helpful in situations such as calling deciders from propulsion as propulsion handlers are now Task based while deciders are still Async.

### DIFF
--- a/src/Equinox/Decider.fs
+++ b/src/Equinox/Decider.fs
@@ -107,6 +107,38 @@ type Decider<'event, 'state>(inner : DeciderCore<'event, 'state>) =
         let! ct = Async.CancellationToken
         let inline decide' c ct = task { let! r, es = Async.StartImmediateAsTask(decide c, ct) in return struct (r, Seq.ofList es) }
         return! inner.TransactExAsync(decide = decide', mapResult = mapResult, ?load = load, ?attempts = attempts, ct = ct) |> Async.AwaitTaskCorrect }
+/// 1. Invoke the supplied <c>Async</c> <c>interpret</c> function with the present state
+    /// 2. (if events yielded) Attempt to sync the yielded events to the stream.
+    ///    (Restarts up to <c>maxAttempts</c> times with updated state per attempt, throwing <c>MaxResyncsExhaustedException</c> on failure of final attempt.)
+    /// 3. Uses <c>render</c> to generate a 'view from the persisted final state
+    member _.TransactTask(interpret : 'state -> CancellationToken -> Task<'event list>, render : 'state -> 'view, ?load, ?attempts, ?ct) : Task<'view> =  task {
+        let inline interpret' s ct = task { let! es = interpret s ct in return Seq.ofList es }
+        return! inner.TransactAsync(interpret', render, ?load = load, ?attempts = attempts, ?ct = ct) }
+
+    /// 1. Invoke the supplied <c>Async</c> <c>decide</c> function with the present state, holding the <c>'result</c>
+    /// 2. (if events yielded) Attempt to sync the yielded events to the stream.
+    ///    (Restarts up to <c>maxAttempts</c> times with updated state per attempt, throwing <c>MaxResyncsExhaustedException</c> on failure of final attempt.)
+    /// 3. Yield result
+    member _.TransactTask(decide : 'state -> CancellationToken -> Task<'result * 'event list>, ?load, ?attempts, ?ct) : Task<'result> =  task {
+        let inline decide' s ct = task { let! r, es = decide s ct in return struct (r, Seq.ofList es) }
+        return! inner.TransactAsync(decide = decide', ?load = load, ?attempts = attempts, ?ct = ct) }
+
+    /// 1. Invoke the supplied <c>Async</c> <c>decide</c> function with the current complete context, holding the <c>'result</c>
+    /// 2. (if events yielded) Attempt to sync the yielded events to the stream.
+    ///    (Restarts up to <c>maxAttempts</c> times with updated state per attempt, throwing <c>MaxResyncsExhaustedException</c> on failure of final attempt.)
+    /// 3. Yield result
+    member _.TransactExTask(decide : ISyncContext<'state> -> CancellationToken -> Task<'result * 'event list>, ?load, ?attempts, ?ct) : Task<'result> =  task {
+        let decide' c ct = task { let! r, es = decide c ct in return struct (r, Seq.ofList es) }
+        return! inner.TransactExAsync(decide = decide', ?load = load, ?attempts = attempts, ?ct = ct) |> Async.AwaitTaskCorrect }
+
+    /// 1. Invoke the supplied <c>Async</c> <c>decide</c> function with the current complete context, holding the <c>'result</c>
+    /// 2. (if events yielded) Attempt to sync the yielded events to the stream.
+    ///    (Restarts up to <c>maxAttempts</c> times with updated state per attempt, throwing <c>MaxResyncsExhaustedException</c> on failure of final attempt.)
+    /// 3. Yields a final 'view produced by <c>mapResult</c> from the <c>'result</c> and/or the final persisted <c>ISyncContext</c>
+    member _.TransactExTask(decide : ISyncContext<'state> -> CancellationToken -> Task<'result * 'event list>, mapResult : 'result -> ISyncContext<'state> -> 'view,
+                             ?load, ?attempts, ?ct) : Task<'view> =  task {
+        let inline decide' c ct = task { let! r, es = decide c ct in return struct (r, Seq.ofList es) }
+        return! inner.TransactExAsync(decide = decide', mapResult = mapResult, ?load = load, ?attempts = attempts, ?ct = ct) }
 
 /// Central Application-facing API. Wraps the handling of decision or query flows in a manner that is store agnostic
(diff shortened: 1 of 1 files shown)


src/Equinox/Decider.fs
```diff
@@ -108,6 +108,39 @@ type Decider<'event, 'state>(inner : DeciderCore<'event, 'state>) =
         let inline decide' c ct = task { let! r, es = Async.StartImmediateAsTask(decide c, ct) in return struct (r, Seq.ofList es) }
         return! inner.TransactExAsync(decide = decide', mapResult = mapResult, ?load = load, ?attempts = attempts, ct = ct) |> Async.AwaitTaskCorrect }
 
+    /// 1. Invoke the supplied <c>Task</c> returning <c>interpret</c> function with the present state
+    /// 2. (if events yielded) Attempt to sync the yielded events to the stream.
+    ///    (Restarts up to <c>maxAttempts</c> times with updated state per attempt, throwing <c>MaxResyncsExhaustedException</c> on failure of final attempt.)
+    /// 3. Uses <c>render</c> to generate a 'view from the persisted final state
+    member _.TransactTask(interpret : 'state -> CancellationToken -> Task<'event list>, render : 'state -> 'view, ?load, ?attempts, ?ct) : Task<'view> =  task {
```
bartelink, line 115: Have you experimented with other signatures? e.g. would calling it `TransactAsync` work or does it mess up intellisense? (I can see the reasoning about giving specific names - we can always add aliases later but you can't undo an overloaded name as easily)
is the `list` -> `seq` hop still valuable (IIRC in the async case, it mattered somewhere at some time, but that might have been an old compiler/VS toolset on net45 a long time ago)

src/Equinox/Decider.fs
```diff
@@ -108,6 +108,39 @@ type Decider<'event, 'state>(inner : DeciderCore<'event, 'state>) =
...
+        let inline interpret' s ct = task { let! es = interpret s ct in return Seq.ofList es }
+        return! inner.TransactAsync(interpret', render, ?load = load, ?attempts = attempts, ?ct = ct) }
+
+    /// 1. Invoke the supplied <c>Task</c> returning <c>decide</c> function with the present state, holding the <c>'result</c>
+    /// 2. (if events yielded) Attempt to sync the yielded events to the stream.
+    ///    (Restarts up to <c>maxAttempts</c> times with updated state per attempt, throwing <c>MaxResyncsExhaustedException</c> on failure of final attempt.)
+    /// 3. Yield result
+    member _.TransactTask(decide : 'state -> CancellationToken -> Task<'result * 'event list>, ?load, ?attempts, ?ct) : Task<'result> =  task {
```
bartelink, line 123: what would you think about forcing `decide` to return struct tuples? and maybe `seq` ?

src/Equinox/Decider.fs
bartelink: Also wondering whether exposing a lower level TransactExAsync (or `inner`) might allow this overall need to be served by adding extension methods/type augmentations locally vs requiring every impl to be a PR and release?

src/Equinox/Decider.fs
```diff
@@ -108,6 +108,39 @@ type Decider<'event, 'state>(inner : DeciderCore<'event, 'state>) =
...
+    member _.TransactExTask(decide : ISyncContext<'state> -> CancellationToken -> Task<'result * 'event list>, ?load, ?attempts, ?ct) : Task<'result> =  task {
+        let decide' c ct = task { let! r, es = decide c ct in return struct (r, Seq.ofList es) }
+        return! inner.TransactExAsync(decide = decide', ?load = load, ?attempts = attempts, ?ct = ct) }
+
+    /// 1. Invoke the supplied <c>Task</c> returning <c>decide</c> function with the current complete context, holding the <c>'result</c>
+    /// 2. (if events yielded) Attempt to sync the yielded events to the stream.
+    ///    (Restarts up to <c>maxAttempts</c> times with updated state per attempt, throwing <c>MaxResyncsExhaustedException</c> on failure of final attempt.)
+    /// 3. Yields a final 'view produced by <c>mapResult</c> from the <c>'result</c> and/or the final persisted <c>ISyncContext</c>
+    member _.TransactExTask(decide : ISyncContext<'state> -> CancellationToken -> Task<'result * 'event list>, mapResult : 'result -> ISyncContext<'state> -> 'view,
+                             ?load, ?attempts, ?ct) : Task<'view> =  task {
```
bartelink, line 140: remove 2 spaces ;)

src/Equinox/Decider.fs
bartelink: Also wondering are any of these signatures useful in dotnet-templates?

bartelink:
The Task vs Async naming feels a bit wonky on first look, even though it's following the logic of the naming in general. Other questions spring to mind:
- should `ct` args be mandatory ? In general the reason I didn't just expose both on a single abstraction is that I wanted people to think before giving up the flowing of cancellation - in general given a Transact will involve an IO roundtrip, the perf overhead of Async vs Task is not going to be significant.
- is forcing `list` and using non-`struct` tuples still appropriate if going `task`?
- is this best addressed by exposing a single low level method and/or `Inner` and then letting people build extension methods? (I'm open to pushback that all you're doing is adding equivalent `Task` APIs for any `Async` ones, but part of me feels we should validate that need in multiple systems before baking it)


I do appreciate the overall idea of removing technical noise from business logic of course, but perhaps having some helpers in a ubiquitous lower level shims lib might address that more generally e.g. https://github.com/fsprojects/FSharp.Control.TaskSeq/issues/142

src/Equinox/Decider.fs
nordfjord: > Have you experimented with other signatures?

I'll admit that I didn't, I've been knee deep in OCaml (which doesn't support any overloading whatsoever) for the past few days so it didn't even occur to me to overload.

If it is possible would that be your preference?

nordfjord: > is the list -> seq hop still valuable

I'm not sure to be honest, I assumed it was there intentionally.

src/Equinox/Decider.fs
nordfjord: > `decide` to return struct tuples?

I'd love it if F# removed the need to add the `struct` keyword from the calling side (does it do that yet?). I've noticed the struct tuple / tuple divide cause confusion in my team at least. Though yes, I do agree that a struct tuple is appropriate here. I'm torn 🤔 

> and maybe `seq`?

I think that'd go against the programming model. We encourage users to set up deciders in a certain way that uses lists, and as such I don't think moving the `seq` transformation to the client provides a lot of value.

src/Equinox/Decider.fs
nordfjord: 😂, even my fancy vim indent config can't handle your demands!

src/Equinox/Decider.fs
bartelink: The root issue here is that we need to know if there are events pretty much straight away in https://github.com/jet/equinox/blob/master/src/Equinox/Core.fs#L44
That was once boring F# async code, and `list` made sense - pattern matching etc
Recently for C#/DeciderCore, I made that be `Task<_, 'res * 'e seq>` as a LCD between C# `yield return`s (which IIRC can't be used to build arrays neatly?) and the `list`s from F#
I suspect that changing the `list` to `seq` will trigger some compiler errors in F# consumption in some relatively exotic cases. (I think I validated that, but not 100% sure)

nordfjord: Isn't a list already a `seq`? can't we just pass the list?

bartelink: Yes, it is. There are exotic cases (e.g. `>>` etc) wherein things sometimes break down though. They are definitely pretty rare and/or arguably shortcomings in type inference and/or type generalization

src/Equinox/Decider.fs
bartelink: there's a pile of `. =` sequences, can you clean them too while you're a it please?

bartelink:
With the latest changes, I wonder whether doing a `member val Core = inner` would not give you pretty much the same experience? or you could call it `Task` or `Internal`. Either way, providing that would give an escape hatch for concocting arbitrary variants by calling `TransactExAsync` on that?

Are you using something like IcedTasks for `CancellationToken -> Task<'T>`? If not, would skipping passing the CT match the real world usage in your app? Obviously one would not expose that in the real API, but having customised helpers live as type augmentations in a local helper lib might be a route to customising that (and hiding threading of cancellation)

bartelink:
>We're moving towards using Tasks more and more and this API would be helpful in situations such as calling deciders from propulsion as propulsion handlers are now Task based while deciders are still Async.

The next RC of Propulsion will provide a lower level API based on Tasks, Funcs and struct tuples
However the default API (without `Async` suffices in `Factory`) will continue to use Async, obj tuples, and FSharpFuncs
Having _ct and/or `Async.AwaitTask` in prod code can wait until a profiler tells you which ones need it

For this release, I'm going to expose a `val Core : DeciderCore` so you can further prototype to cater for such needs by having type augmentations on `Decider` in a global namespace within an app.

As discussed out of band, the key issue with the APIs in DeciderCore from a F# perspective is that they operate on `Event seq` rather than `Event list` (that and the struct tuples)

I'm hopeful that taking a bit longer before baking something noisy in will ultimately yield a better overall result, even if it's not entirely clear to either of us what that might actually become.

Issue #376, opened 2023-04-13
DynamoDB: EventsContext fails to read when tip not read


epNickColeman:
The DynamoStore EventsContext fails to read events when the tip document is not included in any of the batches read.

Example stack trace:
```
 EqxDynamo QueryF ModuleSummary-caded21863a142a9bead26c92ed437d3 v977 397.6ms 95RU 32e/1 9805b >945 <null 
{
    "batchSize": 32
}

W DynamoStoreSource read failure 
{}

System.AggregateException: One or more errors occurred. (unexpected)
---> System.Exception: unexpected
at Equinox.DynamoStore.Core.Query.toPosition(LoadRes _arg1) in /_//src/Equinox.DynamoStore/DynamoStore.fs:line 850
at Equinox.DynamoStore.Core.Query.load@859.MoveNext() in /_//src/Equinox.DynamoStore/DynamoStore.fs:line 882
at <StartupCode$Equinox-DynamoStore>.$DynamoStore.Read@1072.MoveNext()
at <StartupCode$Equinox-DynamoStore>.$DynamoStore.GetInternal@1437.MoveNext() in /_//src/Equinox.DynamoStore/DynamoStore.fs:line 1448
at <StartupCode$Equinox-DynamoStore>.$DynamoStore.clo@1422-3.MoveNext() in /_//src/Equinox.DynamoStore/DynamoStore.fs:line 1423
at Propulsion.DynamoStore.Impl.materializeSpans@71-1.MoveNext() in /_//src/Propulsion.DynamoStore/DynamoStoreSource.fs:line 79
at Propulsion.DynamoStore.Impl.materializeIndexEpochAsBatchesOfStreamEvents@46.MoveNext() in /_//src/Propulsion.DynamoStore/DynamoStoreSource.fs:line 111
at FSharp.Control.TaskSeq`2.System.Threading.Tasks.Sources.IValueTaskSource<System.Boolean>.GetResult(Int16 token)
at System.Threading.Tasks.ValueTask`1.ValueTaskSourceAsTask.<>c.<.cctor>b__4_0(Object state)
--- End of inner exception stack trace ---
at <StartupCode$Propulsion-Feed>.$FeedSource.clo@268-60.MoveNext() in /_//src/Propulsion.Feed/FeedSource.fs:line 272
[...]

PR #377, opened 2023-04-13
DynamoDB: EventContext fails when tip not read

epNickColeman:
Add failing test when reading events from DynamoStore when not reading the batch containing the tip #376

### DIFF
--- a/src/Equinox.DynamoStore/DynamoStore.fs
+++ b/src/Equinox.DynamoStore/DynamoStore.fs
@@ -507,6 +507,7 @@ module internal Position =
     let toVersionAndStreamBytes = function Some p -> p.index, p.calvedBytes + p.baseBytes | None -> 0, 0
     let null_ i = { index = i; etag = null; calvedBytes = 0; baseBytes = 0; unfoldsBytes = 0; events = Array.empty }
     let flatten = function Some p -> p | None -> null_ 0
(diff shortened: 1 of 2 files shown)


bartelink:
Thanks, this helped a lot, much appreciated

Can you pull the upstream and cherry-pick 0d1844a180017172aecf8e0d7da3434b22ef1b1d please?

A lot of the code is as it is as a result of a mix of:
- inherited logic from Cosmos (the append logic is not wired up atm)
- rolling changes (e.g. I did a pass to switch to `task` and `taskSeq` - there was noise due to that)
- matching the Cosmos APIs, so the integration tests can be borrowed too

In this case I'm taking the view that the EventsContext APIs are internal and should be `task`-based, and are internal. As such, I think changing the signatures as I've done here makes sense.

There's definitely scope for follow-on work here e.g.:
- Target the same test for Cosmos too
- cover more scenarios where events have been pruned and fallbacks are required
- make the `toPosition` be more fail-fast in nature 

But for now, this adds important coverage and fixes the issue so we can worry about that another time. If you cherry pick, I'll close this and merge yours before I do the next release (I got #378 in and @nordfjord has https://github.com/jet/equinox/pull/375 pending too)

epNickColeman:
Sorry didn't get back to this till now

bartelink:
Thanks again for the legwork and patience on this

PR #380, opened 2023-05-02
Converge Cache logic

bartelink:
Precursor to #381, standardising the caching logic in order to be able to introduce the cache validation cleanly


nordfjord (approved):
Had a quick scroll through, looks good to me, love that the caching logic is becoming core

src/Equinox.Core/CachingDecorator.fs
```diff
@@ -0,0 +1,51 @@
...
+open System.Threading
+open System.Threading.Tasks
+
+type IReloadableCategory<'event, 'state, 'context> =
+    inherit ICategory<'event, 'state, 'context>
+
+    abstract Reload: log: ILogger * streamName: string * requireLeader: bool * streamToken: StreamToken * state: 'state * ct: CancellationToken
+                     -> Task<struct (StreamToken * 'state)>
+
+type CachingDecorator<'event, 'state, 'context>
```
nordfjord, line 14: This is a great change! Did you notice any deviations between the "caching category" implementations when extracting this, or was this very clean cut?

src/Equinox.Core/CachingDecorator.fs
```diff
@@ -0,0 +1,51 @@
...
+
+open Serilog
+open System
+open System.Threading
+open System.Threading.Tasks
+
+type IReloadableCategory<'event, 'state, 'context> =
+    inherit ICategory<'event, 'state, 'context>
+
+    abstract Reload: log: ILogger * streamName: string * requireLeader: bool * streamToken: StreamToken * state: 'state * ct: CancellationToken
```
nordfjord, line 11: Aren't all categories reloadable? can we avoid adding another interface here? Reloading could default to returning the current token and state, or the empty token and state

nordfjord: Or is the point more that the stores should implement reloadable categories, while consumers of the library deal with ICategories?

src/Equinox.DynamoStore/Equinox.DynamoStore.fsproj
```diff
@@ -5,6 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Compile Include="..\Equinox.Core\CachingDecorator.fs" Link="CachingDecorator.fs" />
```
nordfjord, line 8: We already depend on Equinox Core in this Project, so this should be unnecessary

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -469,23 +448,19 @@ type MessageDbCategory<'event, 'state, 'context>(resolveInner, empty) =
...
-            | Some (CachingStrategy.FixedTimeSpan (cache, period)) ->
-                Caching.applyCacheUpdatesWithFixedTimeSpan cache null period folder Token.supersedes
-            | Some (CachingStrategy.SlidingWindowPrefixed (cache, window, prefix)) ->
-                Caching.applyCacheUpdatesWithSlidingExpiration cache prefix window folder Token.supersedes
-        let resolveInner categoryName streamId = struct (category, StreamName.render categoryName streamId, ValueNone)
+            | None -> (fun _ -> Task.FromResult ValueNone), fun _ _ -> Task.FromResult ()
+            | Some (CachingStrategy.SlidingWindow (cache, window)) -> tryGet cache, Cache.updateWithSlidingExpiration (cache, null) window Token.supersedes
+            | Some (CachingStrategy.FixedTimeSpan (cache, period)) -> tryGet cache, Cache.updateWithFixedTimeSpan (cache, null) period Token.supersedes
+            | Some (CachingStrategy.SlidingWindowPrefixed (cache, window, prefix)) -> tryGet cache, Cache.updateWithSlidingExpiration (cache, prefix) window Token.supersedes
+        let cached = CachingDecorator<'event, 'state, 'context>(raw, tryReadCache, updateCache) : ICategory<_, _, _>
```
nordfjord, line 459: slight difference in behaviour vs what was here previously. Previously we'd not instantiate the wrapper if no caching was desired.

Not the biggest of deals, a couple of wasted cycles per call, just thought I'd call it out

src/Equinox.Core/CachingDecorator.fs
nordfjord: Alright, read through the rest of the code, looks like all the stores primarily implement this concept rather than the ICategory directly. I wonder if we should make this definition stand on its own rather than inherit from ICategory since users will never see this type anyway and inheriting, while powerful, gives me the shivers

nordfjord (approved):
none of the above comments are blockers from my perspective

src/Equinox.DynamoStore/Equinox.DynamoStore.fsproj
bartelink: Yeah, but it's `None` in the Core fsproj (and the helpers and types are/were intended to be internal). Main thing is that `Core.Caching` would be in the API/intellisense surface.
But yes, will likely remove this messing as the MemoryStore is the only store that does not use it anyway

src/Equinox.Core/CachingDecorator.fs
bartelink: While there were once some minor arbitrary variations, I had been gently nudging them to convergence over time, so the high level logic was identical

src/Equinox.Core/CachingDecorator.fs
bartelink: Good point that I'll take away and ponder...

You are correct that MemoryStore could reasonably implement it too, and it's a pretty central part of any reasonably efficient even sourced infra (i.e. a reload of kinds needs to happen in the case of a resync even if you are not doing caching), and/or provide a null impl
A big reason that the file is not compiled into Core is that this is just an impl detail.
It may also be possible and/or reasonable to twist things such that the CachingStrategy DU resides in Core and the impls don't need the duplicated mapping/wiring.

I had been resisting having a big Core lib from the very start - being able to point people to 2/3 files: https://github.com/jet/equinox/blob/master/src/Equinox/Core.fs + https://github.com/jet/equinox/blob/master/src/Equinox.MemoryStore/MemoryStore.fs and saying "that's all it is" is a key value prop

Stuff like 
- the prohibition on caching with EventsAreState Mode https://github.com/jet/equinox/blob/doc-tweaks/src/Equinox.EventStoreDb/EventStoreDb.fs#L446-L451 (I couldn't think of a cleaner way to wire it at the time - it's not a critical design masterplan)
- the mandatory/explicit caching mode for Cosmos/Dynamo (having to say NoCaching vs not specifying a Caching in the other stores)
becomes more debatable if making specifying the caching can simply be made orthogonal

src/Equinox.MessageDb/MessageDb.fs
bartelink: Hm yeah it should be possible to restore the previous logic by wrapping the instantiation of the Decorator in a small helper and having the preceding match yield the decorated form or the raw one depending on whether it's needed

src/Equinox.DynamoStore/Equinox.DynamoStore.fsproj
bartelink: Caching.fs is now in Core

src/Equinox.MessageDb/MessageDb.fs
bartelink: null wrapping now removed in Caching.fs

src/Equinox.Core/CachingDecorator.fs
bartelink: Merging...
- MemoryStore is not currently reloadable - still on the fence about whether it's important that the memorystore be minimal and demonstrates caching is optional
- open to a PR to make `IReloadable` a separate interface not derived from `ICategory`
- ICategory is prob a bad name

PR #386, opened 2023-05-15
Cache: Add ReadThrough mode

### LINKED ISSUE #381: Feature: ReadThrough mode
The current AllowStale option will use any cached state as held

Being able to constrain it to only use the state if it's within a certain age range will enable users to trigger periodic re-reads to freshen the state

To facilitate heavy usage as a read-through cache, reads under this mode should be concentrated internally via an `AsyncCacheCell` in order to ensure a timeout does not read to a flurry of concurrent update attempts

NOTE just letting the underlying entry in the MemoryCache expire is insufficient - we still need to have the entry be usable as base state for the read (e.g. in order to allow etag checked reads) and to be able to use it to control the single reader gating

bartelink:
This makes `decider.Query(..., load = Equinox.AllowStale (TimeSpan.FromSeconds 1)` opt into:
1. have all reads where the cache entry is stale or missing share a single roundtrip to load the state from the store
2. cache the value for the specified time (from the start of the retrieval, not when it arrived)

This guarantees that the effects of processing reads against a stream whose core function is to manage the write side are limited
- the readers within a given process will self-limit themselves to maximum one roundtrip per specified period
- if a writer updates in the interim (using the same `Cache`), ReadThrough reads can reuse that value, avoiding a retrieval roundtrip

Open questions:
- [x] ~Should `Cache` implement `IDisposable` (it will trigger warnings in client mode as `Cache` construction will have to be called via `new Cache`)~ No, people don't Dispose Caches IRL is the position for now...
- [x] ~Should caching of state be made possible for MemoryStore too?~ Nah, not this time.
- [x] Should `AllowStale` become `AnyCachedVersion` given the fact that less surprising semantics are now available? (should be considered alongside naming of `MaxStaleness`)
- [x] Rename `MaxStaleness` to `AllowStale` (means compiler errors for people leaning on the older behavior that's now entitled `AnyCachedValue`, which I believe is a good tradeoff on balance)

See also #388 and #380 for prep/tidy work on the way to this
Resolves #381


src/Equinox.Core/Caching.fs
```diff
@@ -7,62 +7,52 @@ open System.Diagnostics
...
+type IReloadable<'state> =
     abstract Reload: log: ILogger * streamName: string * requireLeader: bool * streamToken: StreamToken * state: 'state * ct: CancellationToken
                      -> Task<struct (StreamToken * 'state)>
 
-type private Decorator<'event, 'state, 'context>
-    (   category: IReloadableCategory<'event, 'state, 'context>,
-        tryReadCache: string -> Task<voption<struct (StreamToken * 'state)>>,
-        updateCache: string -> struct (StreamToken * 'state) -> Task<unit>) =
-    let cache streamName (inner: CancellationToken -> Task<struct (StreamToken * 'state)>) ct = task {
+type private Decorator<'event, 'state, 'context, 'cat when 'cat :> ICategory<'event, 'state, 'context> and 'cat :> IReloadable<'state> >
```
bartelink, line 14: @nordfjord this was what I was trying to do in a rush the other night, based on your distaste of hierarchy. As the PR as a whole has a slight conflict with #387, feel free to review this, and I'll cherry pick this to master once you've looked, with a view to releasing that when you're ready

bartelink:
@DSilence The `ICache` interface changes significantly, and is obviously harder to implement now, as it does more.
I don't hate it, but have to say I am tempted to remove `ICache` as I'm not convinced about there being meaningful implementations out there

Though it also has to be said that these series of PRs (esp the cleanup ones I linked above) are the realisation of a lot of stuff you were pushing for at the time you introduced them (recall I didn't want to have a common cache impl until it REALLY hurt and then figure out the best design at that point)

src/Equinox.Core/Cache.fs
```diff
@@ -41,34 +44,87 @@ module private CachingStrategy =
...
-            let freshEntry = CacheEntry(token, state)
-            match inner.AddOrGetExisting(key, freshEntry, CachingStrategy.toPolicy options) with
-            | null -> Task.FromResult()
-            | :? CacheEntry<'state> as existingEntry -> existingEntry.UpdateIfNewer(compare, freshEntry); Task.FromResult()
-            | x -> failwithf "UpdateIfNewer Incompatible cache entry %A" x
+        // if there's a non-zero maxStaleness, concurrent read attempts share the roundtrip (and its fate, if it throws)
+        member x.Load(key, maxStaleness, compare, options, loadOrReload) = task {
+            let loadOrReload maybeBaseState () =
+                let act = System.Diagnostics.Activity.Current
+                if act <> null then act.AddCacheHit(ValueOption.isSome maybeBaseState) |> ignore
```
nordfjord, line 116: 🙏

bartelink: BTW not certain whether there's value in separated metrics for incremental loads due to caching vs AllowStale roundtrips saved, or whether a cache hit is the main point - will be pondering

PR #388, opened 2023-05-18
Prepare for ReadThrough mode

bartelink:
#386 focuses on the implementation. This PR generalises surrounding functionality in preparation for this
- fixes some stale comments re caching
- extract a formal `IReloadable` interface for the cache to interact with
- generalises the `supersedes` store function used to compare cache entries to a `compare` function that can also confirm an entry is identical (in order to be able to track when a cache entry has been successfully validated)
- generalises the `allowStale` flag to be a `maxStaleness: TimeSpan` (`true`/`false` becomes `TimeSpan.MaxValue`/`TimeSpan.Zero`)
- simplifies/clarifies the AsyncCacheCell internals to align with the requirements of the ReadThrough mode implementation
- makes the `CacheEntry` type internal
- adds baseline tests for the Caching behavior (in a long overdue `Equinox.Core.Tests` project)
- removes dead code


src/Equinox.Core/Caching.fs
```diff
@@ -3,66 +3,59 @@ module Equinox.Core.Caching
...
+        let act = System.Diagnostics.Activity.Current
+        if act <> null then act.AddCacheHit(ValueOption.isSome cacheItem) |> ignore
+        return cacheItem }
+    let save key (inner: CancellationToken -> Task<struct (StreamToken * 'state)>) ct = task {
+        let! struct (token, state) as res = inner ct
+        do! cache.UpdateIfNewer(key, compareTokens, createOptions (), token, state)
+        return res }
     interface ICategory<'event, 'state, 'context> with
-        member _.Load(log, categoryName, streamId, streamName, allowStale, requireLeader, ct) = task {
-            match! tryReadCache streamName with
```
bartelink, line 26: note this was not computing a correct key in the case of SlidingWindowPrefixed (fixed in #386)

src/Equinox/Core.fs
```diff
@@ -13,7 +13,7 @@ type IStream<'event, 'state> =
     abstract LoadEmpty: unit -> struct (StreamToken * 'state)
 
     /// Obtain the state from the target stream
-    abstract Load: allowStale: bool * requireLeader: bool * ct: CancellationToken -> Task<struct (StreamToken * 'state)>
+    abstract Load: maxStaleness: TimeSpan * requireLeader: bool * ct: CancellationToken -> Task<struct (StreamToken * 'state)>
```
nordfjord, line 16: Suggestion: `maxStaleness` -> `maxAge`

src/Equinox.DynamoStore/DynamoStore.fs
```diff
@@ -1131,11 +1131,9 @@ type internal Category<'event, 'state, 'context>
         match! store.Reload(log, (streamNam, pos), requireLeader, (codec.TryDecode, isOrigin), ct) with
         | LoadFromTokenResult.Unchanged -> return struct (streamToken, state)
         | LoadFromTokenResult.Found (token', events) -> return token', fold state events }
-    interface Caching.IReloadableCategory<'event, 'state, 'context> with
-        member _.Load(log, _categoryName, _streamId, streamName, _allowStale, requireLeader, ct) =
+    interface ICategory<'event, 'state, 'context> with
+        member _.Load(log, _categoryName, _streamId, streamName, _maxStaleness, requireLeader, ct) =
```
nordfjord, line 1135: This is probably pragmatic, but seeing that every store implementation is ignoring the `_maxStaleness` parameter has me wondering if this is the right interface 🤔

src/Equinox/Decider.fs
```diff
@@ -244,9 +244,9 @@ and internal LoadPolicy() =
     static member Fetch<'state, 'event>(x: LoadOption<'state> option)
         : IStream<'event, 'state> -> CancellationToken -> Task<struct (StreamToken * 'state)> =
         match x with
-        | None | Some RequireLoad ->                 fun stream ct ->   stream.Load(allowStale = false, requireLeader = false, ct = ct)
-        | Some RequireLeader ->                      fun stream ct ->   stream.Load(allowStale = false, requireLeader = true,  ct = ct)
-        | Some AllowStale ->                         fun stream ct ->   stream.Load(allowStale = true,  requireLeader = false, ct = ct)
+        | None | Some RequireLoad ->                 fun stream ct ->   stream.Load(maxStaleness = TimeSpan.Zero,     requireLeader = false, ct = ct)
+        | Some RequireLeader ->                      fun stream ct ->   stream.Load(maxStaleness = TimeSpan.Zero,     requireLeader = true,  ct = ct)
+        | Some AllowStale ->                         fun stream ct ->   stream.Load(maxStaleness = TimeSpan.MaxValue, requireLeader = false, ct = ct)
```
nordfjord, line 249: Nit: in theory a system could be running beyond the heat death of the universe and get a stale value in this branch 🤡

src/Equinox/Core.fs
bartelink: Good call, done (still using that as the name for the mode, but #386 already calls that out)

src/Equinox.DynamoStore/DynamoStore.fs
bartelink: The other idea I have in that space is for the layers inward of the cache to have a `LoadOrReload` signature with a `origin: (token*state) voption`, rather than Load+Reload (should that be called `Fetch` ?). ICategory and ICategoryImpl ;)

src/Equinox.Core/Cache.fs
```diff
@@ -1,61 +1,27 @@
...
 
 type ICache =
-    abstract member UpdateIfNewer: key: string * options: CacheItemOptions * (struct (StreamToken * StreamToken) -> bool) * entry: CacheEntry<'state> -> Task<unit>
-    abstract member TryGet: key: string -> Task<struct (StreamToken * 'state) voption>
+    abstract member TryGet:
+                          key: string
+                           -> System.Threading.Tasks.Task<struct (StreamToken * 'state) voption>
+    abstract member UpdateIfNewer:
+                          key: string
+                          * compareTokens:(struct (StreamToken * StreamToken) -> int64)
```
nordfjord, line 13: 🤔 could this be `StreamToken -> StreamToken -> StreamToken`?

src/Equinox/Decider.fs
bartelink: not unrelated, I was pondering whether the etag in DynamoStore should just become an int64 update count - would save lots of allocations to boot

nordfjord: That makes total sense to me

src/Equinox.Core/Cache.fs
bartelink: YES, thank you - it's one unnecessarily ugly duck as it is
But it's a `struct` (and prob should stay that), so the caller would need to do a `ReferenceEquals` on `.value` to be able to identify when a superseded value was ignored.
I guess `-> StreamToken voption` can convey that more cleanly

src/Equinox/Decider.fs
bartelink: >That makes total sense to me
Running? Wanting to break the format for DynamoStore and/or write tests to validate that an impl has backcompatiblity? :D
If only there were phalanxes of contributors ready to pounce on up-for-grabs issues for the kudos IRL :D

src/Equinox.Core/Cache.fs
bartelink: went with isStale  - it's in the negative, but is meaningful to someone scanning (versus `isValid`), thanks for prompting!

PR #389, opened 2023-05-18
Remove ICache

bartelink:
Spike - it could be this much shorter if we removed `ICache` ;) @DSilence


nordfjord (approved):
YOLO 🚀 

Jokes aside, I'm not aware of another implementation of ICache anywhere. At least not out in the open https://github.com/search?type=code&q=%22interface+ICache+with%22+language%3AF%23

bartelink:
Yeah, @DSilence put it in for a reason though - the company he was working for was using another cache (which had Task interfaces which is why there was egregious Asyncness in the interface).
But it was the V3 rollover last time I asked him, so maybe he'll be all YOLO on it for V4 :D

bartelink:
@DSilence Thinking I'll bite the bullet on this as it makes the code more legible and removes a concept - if the system in question is upgrading, we'll hopefully be able to work out a solution without impeding that (i.e. open to reintroducing at some point, but taking the lack of traffic here as assent!)

PR #387, opened 2023-05-16
MessageDb: cleanup trace spans hierarchy

nordfjord:
Screenshot 2023-05-15 at 23 08 42

It's bothering me that we essentially create three spans to represent a single operation. With this change I'm hoping to bring it down to 2, relying on the Npgsql instrumentation to see what is actually going on in the database

PR #393, opened 2023-06-06
MessageDb: Condense trace - fewer spans

nordfjord:


bartelink:
Thanks, looks good. Will leave it here in case you have follow-ups until just before I release (in the coming days)

bartelink:
related: #387

PR #394, opened 2023-06-11
perf: use a literal for queries, and stop doing silly stuff with correlation/causation ids

nordfjord:
Noticed during my memory leak investigation in propulsion that using `@Variable` style Commands in Npgsql leads to `StringBuilder` allocations and all sorts of fun processing in the application that is avoidable by using `$1, $2` etc (the native format that pg parameterised queries take).

Secondly while scanning the code the queries looked a bit off with the `CorrelationId` and `CausationId` being picked from the json metadata. This makes sense if and only if we also place the id's onto these fields during writes. Since we don't this is just waste so 🧹  it

### DIFF
--- a/src/Equinox.MessageDb/MessageDbClient.fs
+++ b/src/Equinox.MessageDb/MessageDbClient.fs
@@ -17,18 +17,18 @@ type ExpectedVersion = Any | StreamVersion of int64
 [<AutoOpen>]
 module private Sql =
     type private NpgsqlParameterCollection with
-      member this.AddNullableString(value: string option) =
-          match value with
-          | Some value -> this.AddWithValue(NpgsqlDbType.Text, value)
-          | None       -> this.AddWithValue(NpgsqlDbType.Text, DBNull.Value)
-      member this.AddExpectedVersion(value: ExpectedVersion) =
-        match value with
-        | StreamVersion value -> this.AddWithValue(NpgsqlDbType.Bigint, value)
-        | Any                 -> this.AddWithValue(NpgsqlDbType.Bigint, DBNull.Value)
+        member this.AddNullableString(value: string option) =
+            match value with
+            | Some value -> this.AddWithValue(NpgsqlDbType.Text, value)
+            | None       -> this.AddWithValue(NpgsqlDbType.Text, DBNull.Value)
+        member this.AddExpectedVersion(value: ExpectedVersion) =
+            match value with
+            | StreamVersion value -> this.AddWithValue(NpgsqlDbType.Bigint, value)
+            | Any                 -> this.AddWithValue(NpgsqlDbType.Bigint, DBNull.Value)
 
-      member this.AddJson(value: Format) =
-        if value.Length = 0 then this.AddWithValue(NpgsqlDbType.Jsonb, DBNull.Value)
-        else this.AddWithValue(NpgsqlDbType.Jsonb, value.ToArray())
+        member this.AddJson(value: Format) =
+            if value.Length = 0 then this.AddWithValue(NpgsqlDbType.Jsonb, DBNull.Value)
+            else this.AddWithValue(NpgsqlDbType.Jsonb, value.ToArray())
 
 
 module private Queries =
--- a/src/Equinox.MessageDb/MessageDbClient.fs
+++ b/src/Equinox.MessageDb/MessageDbClient.fs
@@ -14,15 +14,24 @@ type private Format = ReadOnlyMemory<byte>
 [<Struct>]
 type ExpectedVersion = Any | StreamVersion of int64
 
+module private Queries =
+    [<Literal>]
+    let writeMessage = "select * from write_message($1::text, $2, $3, $4, $5, $6)"
+    [<Literal>]
+    let readLast = "select position, type, data, metadata, id::uuid, time from get_last_stream_message($1, $2);"
+    [<Literal>]
+    let readStream = "select position, type, data, metadata, id::uuid, time from get_stream_messages($1, $2, $3);"
+
+
 module private Sql =
-    let addExpectedVersion name (value: ExpectedVersion) (p: NpgsqlParameterCollection) =
+    let addExpectedVersion (value: ExpectedVersion) (p: NpgsqlParameterCollection) =
         match value with
-        | StreamVersion value -> p.AddWithValue(name, NpgsqlDbType.Bigint, value) |> ignore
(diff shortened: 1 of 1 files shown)


src/Equinox.MessageDb/MessageDbClient.fs
```diff
@@ -14,15 +14,24 @@ type private Format = ReadOnlyMemory<byte>
 [<Struct>]
 type ExpectedVersion = Any | StreamVersion of int64
 
+module private Queries =
+    [<Literal>]
+    let writeMessage = "select * from write_message($1::text, $2, $3, $4, $5, $6)"
```
bartelink, line 19: 🤔 any advantage in explicitly listing the desired returns?

src/Equinox.MessageDb/MessageDbClient.fs
```diff
@@ -14,15 +14,24 @@ type private Format = ReadOnlyMemory<byte>
...
-        | StreamVersion value -> p.AddWithValue(name, NpgsqlDbType.Bigint, value) |> ignore
-        | Any                 -> p.AddWithValue(name, NpgsqlDbType.Bigint, DBNull.Value) |> ignore
-    let addNullableString name (value: string option) (p: NpgsqlParameterCollection) =
+        | StreamVersion value -> p.AddWithValue(NpgsqlDbType.Bigint, value) |> ignore
+        | Any                 -> p.AddWithValue(NpgsqlDbType.Bigint, DBNull.Value) |> ignore
+    let addNullableString (value: string option) (p: NpgsqlParameterCollection) =
         match value with
-        | Some value -> p.AddWithValue(name, NpgsqlDbType.Text, value) |> ignore
-        | None       -> p.AddWithValue(name, NpgsqlDbType.Text, DBNull.Value) |> ignore
+        | Some value -> p.AddWithValue(NpgsqlDbType.Text, value) |> ignore
```
bartelink, line 33: could just `defaultArg` ?

src/Equinox.MessageDb/MessageDbClient.fs
```diff
@@ -14,15 +14,24 @@ type private Format = ReadOnlyMemory<byte>
 [<Struct>]
 type ExpectedVersion = Any | StreamVersion of int64
 
+module private Queries =
```
bartelink, line 17: I dont see any symmetry here which suggests keeping them together adds value
esp for the write_message, you want the code to be beside the query to show the correlation between the order of adding to the list vs the order in the string, no?

src/Equinox.MessageDb/MessageDbClient.fs
```diff
@@ -14,15 +14,24 @@ type private Format = ReadOnlyMemory<byte>
...
+    let writeMessage = "select * from write_message($1::text, $2, $3, $4, $5, $6)"
+    [<Literal>]
+    let readLast = "select position, type, data, metadata, id::uuid, time from get_last_stream_message($1, $2);"
+    [<Literal>]
+    let readStream = "select position, type, data, metadata, id::uuid, time from get_stream_messages($1, $2, $3);"
+
+
 module private Sql =
-    let addExpectedVersion name (value: ExpectedVersion) (p: NpgsqlParameterCollection) =
+    let addExpectedVersion (value: ExpectedVersion) (p: NpgsqlParameterCollection) =
```
bartelink, line 27: having this off over here doesnt add much value - why not do a one-liner match to illustrate the value mapping, and then do a normal AddWithValue inline (maybe with a `let inline t v (p: ParamCol) -> p.AddWithValue(t, v) |> ignore)

src/Equinox.MessageDb/MessageDbClient.fs
bartelink: Or Maybe convert this to a type augmentation (with autoopen as the module is provate anyway)
type ParamCol with
    member p.AddNullableString(t, v) = p.AddWithValue(t, defaultArg v DBNull.value) |> ignore

src/Equinox.MessageDb/MessageDbClient.fs
nordfjord: I think you're going to have to clarify what you mean here 🤔

src/Equinox.MessageDb/MessageDbClient.fs
nordfjord: > could just defaultArg?

Not quite. DbNull is not a string unfortunately so `defaultArg strOpt DbNull` won't compile

> Or Maybe convert this to a type augmentation (with autoopen as the module is provate anyway)

Done

src/Equinox.MessageDb/MessageDbClient.fs
nordfjord: Main goal here is to make absolutely sure that the string for the query is allocated once and only once. I don't know a good way to do that while also maintaining colocation of the query and parameter application

src/Equinox.MessageDb/MessageDbClient.fs
nordfjord: > why not do a one-liner match to illustrate the value mapping, and then do a normal AddWithValue inline

```fsharp
let expectedVersionToBigInt v = match v with StreamVersion v -> v | Any -> DbNull
```

^ I don't think this would compile

src/Equinox.MessageDb/MessageDbClient.fs
nordfjord: I think I caught your drift, I've colocated the query text and the command preparation

src/Equinox.MessageDb/MessageDbClient.fs
bartelink: wondering if the `*` could/should be replaced with something more selective? (Doesn't really matter - I dont know about stored proc invocation syntax)

nordfjord: Ah, I see now. IIRC the function returns a single bigint representing the new position of the stream (which we already know application side). Not sure if it was a named tuple that you can extract from or not 🤔

src/Equinox.MessageDb/MessageDbClient.fs
bartelink: yes, needs a box (see my push, but doesnt matter)

src/Equinox.MessageDb/MessageDbClient.fs
bartelink: forget I said anything ;)

bartelink:
All very neat, thanks!

PR #395, opened 2023-06-11
Polish Npgsql extensions

bartelink:


bartelink:
Don't know if this actually improves things @nordfjord - releasing without it in 2 mins, as this is not tested (though I have decompiled to verify the transformations are safe as much as possible..)

PR #398, opened 2023-06-12
Rename static DeciderCore -> Factory

bartelink:
Removes an ambiguity the F# compiler chokes on between the factory/extension methods module and the generic DeciderCore type which can require namespace `open`s we'd like to avoid forcing
Example: https://github.com/jet/propulsion/blob/90c843eaa495b440ad1a31223548eaf0f555f0c8/src/Propulsion.EventStore/Checkpoint.fs#L96-L97


bartelink:
Arguably something more like DeciderCore would make sense - the consumption looks [like this](https://github.com/jet/dotnet-templates/blob/b95da5afeb1584593026ff57fbb2ff0e561bc9c3/propulsion-sync/Program.fs#L476) at present

bartelink:
Merged with a bad name; the key gain is we get to remove the ugly disambiguation requirement in Propulsion and/or other places where DeciderCore is used from F#.
Depending on time, @nordfjord might converge Decider+DeciderCore before 4 goes final...

PR #399, opened 2023-06-19
feat(Cosmos, Dynamo): Add CacheStrategy.SlidingWindowPrefixed

bartelink:
Had been holding off on adding this mode, but ran into what I believe represents a relevant use case


bartelink:
Details of use case

When using Read Through mode (#386), some reads may wish to cache values that are derived from Events, and not present in the State. Example cases:
- the relevant state is not used to support decision logic, used only infrequently and/or is too large to include in the snapshot
- the need for the state only becomes apparent after the prod system has lots of events that don't have an in-date snapshot present (and the extra reads would be untenable)

An example might be that a Created event with a large payload that's only read very infrequently, and is not relevant to normal decision making

bartelink:
Abandoned in favor of #417

PR #402, opened 2023-07-03
Avoid null reference on cache read

epNickColeman:
Handle nulls when unpacking Equinox.DynamoStore.Core.Token.isStale

### DIFF
--- a/src/Equinox.DynamoStore/DynamoStore.fs
+++ b/src/Equinox.DynamoStore/DynamoStore.fs
@@ -990,7 +990,10 @@ module Token =
         { value = box { pos = pos }; version = v; streamBytes = b }
     let create: Position -> StreamToken = Some >> create_
     let empty = create_ None
-    let (|Unpack|) (token: StreamToken): Position option = let t = unbox<Token> token.value in t.pos
+    let (|Unpack|) (token: StreamToken): Position option =
+        match token.value  with
+        | :? Token as t -> t.pos
+        | _ -> None
 
     // TOCONSIDER for RollingState, comparing etags is not meaningful, and they are under our control;
     //            => should be replaced with a `revision` that increments if `index` is not changing as part of a write
--- a/tests/Equinox.CosmosStore.Integration/AccessStrategies.fs
+++ b/tests/Equinox.CosmosStore.Integration/AccessStrategies.fs
@@ -1,14 +1,21 @@
 module Equinox.Store.Integration.AccessStrategies
 
 #if STORE_DYNAMO
+open Equinox.Core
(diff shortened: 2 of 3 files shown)


bartelink:
Thanks, not sure how I convinced myself that was not the problem!

tests/Equinox.CosmosStore.Integration/AccessStrategies.fs
```diff
@@ -133,3 +140,33 @@ type UnoptimizedTipReadingCorrectness(testOutputHelper) =
...
+
+
+module Token =
+    #if STORE_DYNAMO
+    let getPos index = { index = index; etag = ""; calvedBytes = 0; baseBytes = 0; unfoldsBytes = 0; events = Array.empty }
+    #else
+    let getPos index = { index = index; etag = None }
+    #endif
+    let [<Fact>] ``Candidate is not stale if we have no current`` () =
+        let emptyToken = Unchecked.defaultof<StreamToken>
```
bartelink, line 152: I wonder if this can be triggered by having L134 have a variant that passes Equinox.LoadOption.AnyCachedValue or similar (but will focus on getting this released instead!)

src/Equinox.DynamoStore/DynamoStore.fs
```diff
@@ -990,7 +990,10 @@ module Token =
         { value = box { pos = pos }; version = v; streamBytes = b }
     let create: Position -> StreamToken = Some >> create_
     let empty = create_ None
-    let (|Unpack|) (token: StreamToken): Position option = let t = unbox<Token> token.value in t.pos
+    let (|Unpack|) (token: StreamToken): Position option =
```
bartelink, line 993: This obviously needs to be applied for the other stores too.
I suspect the cleanest answer might be to have a Null Object a la `TokenEmpty` (which the non-Document Stores have) be put into the token `obj` rather than handling nulls exactly like this, but hard to argue with the coverage as it is!.

bartelink: (A complete solution should feed into the #400 branch)

bartelink:
@epNickColeman The expectation of any valid StreamToken is that the `.value` is non-`null`

On reflection (OK, on eyeballing the code with sufficient sleep in the cold light of day), the technically correct and/or complete thing to do here is to have all `Unpack` implementations validate this precondition, and throw an `ArgumentNullException` if passed an invalid value.

The fault is hence with the caching logic - it's generating invalid `StreamToken` values and then passing those to `isStale`. The test stub impl of isStale was not inspecting the token at all (which is something that should be covered...)

Will adjust this in the context of #403

PR #406, opened 2023-07-20
perf(MessageDb): Fold the stream as we go instead of buffering events

nordfjord:


src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -206,7 +206,7 @@ module Read =
             let batchLog = log |> Log.prop "batchIndex" batchCount
             let! slice = readSlice pos batchCount batchLog ct
             version <- max version slice.LastVersion
-            result.AddRange(slice.Messages)
+            state <- slice.Messages |> Array.chooseV tryDecode |> fold state
```
bartelink, line 209: seq.chooseV - contract does not require it to be an array (was putting it into an array for best memory usage and to be sure the db logic can release stuff etc)

nordfjord: > contract does not require it to be an array

Should it? It'd make the calling code not need the explicit type annotations
```fsharp
let fold = Array.fold evolve
// vs
let fold: State -> Event seq -> State = Seq.fold evolve
```

```suggestion
            state <- slice.Messages |> Seq.chooseV tryDecode |> fold state
```

bartelink: If I wanted to convey the concept that you have the choice of folding them in groups to optimise, the array makes sense

The actual reason to drop it to seq is twofold
- let people do manipulations (upconversions etc, without having to create interstitial arrays)
- I'm not 100% sure but in general in F# its that tiny bit easier to use `list` in tests in some cases, and having the interface be seq makes that more possible

I'll grant you that the Seq.fold evolve syntax thing is a pain (I have not looked for a better solution than the type annotation for years though)
does `let fold s = Seq.fold evolve s` work (not saying it's necessarily better though)?

But there may be swings and roundabouts in this, e.g. there may be common stores that force it to arrays all the time so on balance it might make things simpler - we're not V4 yet if you're up for the spiking (but I would say only 40% chance of a good result)!

nordfjord: > If I wanted to convey the concept that you have the choice of folding them in groups to optimise, the array makes sense

A common pattern I have in my JS applications is

```ts
const fold = (state, events) => events.length ? last(events).data : state
```
For streams where the state can be reconstructed from the latest event. Using a seq would imply this is unwise because you have to iterate the entire seq to get the last element

bartelink: I did once have special case behavior for AnyKnownState mode and went down that rabbit hole
The load will only load 0/one when configured that way, so a `Seq.tryExactlyOne` would be an equivalently pleasing semantic, no?
I think I had some Seq.tryLast stuff too. Can't recall
The job here is to fold though and you should not be called with empty events
Does the fold function accepted by AnyKnownState even give you the state to fold from (it shouldnt)

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -192,12 +192,12 @@ module Read =
             "Read", count, slice.LastVersion)
         return slice }
 
-    let private readBatches (log: ILogger) batchSize (readSlice: int64 -> int -> ILogger -> CancellationToken -> Task<StreamEventsSlice>)
+    let private readBatches (log: ILogger) fold initial tryDecode batchSize (readSlice: int64 -> int -> ILogger -> CancellationToken -> Task<StreamEventsSlice>)
```
bartelink, line 195: initial should be something like origin
batchSize is unused

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -215,17 +215,15 @@ module Read =
...
             let act = Activity.Current
             if act <> null then act.AddBatches(batchCount).AddLastVersion(version) |> ignore
-            return version, Array.ofSeq result }
+            return version, state }
 
-    let private logBatchRead streamName t events (batchSize: int64) version (log: ILogger) =
-        let bytes, count = resolvedEventBytes events, events.Length
-        let reqMetric: Log.Measurement = { stream = streamName; interval = t; bytes = bytes; count = count}
-        let batches = (events.Length - 1)/ (int batchSize) + 1
+    let private logBatchRead streamName t version (log: ILogger) =
```
bartelink, line 220: needs to be passed the event count from the batch read
I also think at least a proxy for the size should be included in a bytes field
This will stll be true if it goes all OT - these things can be very useful for troubleshooting and/or analysis

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -324,22 +322,22 @@ type MessageDbContext(client: MessageDbClient, batchOptions: BatchOptions) =
...
         let! version, events = Read.loadLastEvent log client.ReadRetryPolicy client.Reader requireLeader streamName None ct
-        return Token.create version, Array.chooseV tryDecode events }
+        return struct(Token.create version, events |> Array.chooseV tryDecode |> fold initial) }
     member _.LoadSnapshot(log, category, streamId, requireLeader, tryDecode, eventType, ct) = task {
         let snapshotStream = Snapshot.streamName category streamId
         let! _, events = Read.loadLastEvent log client.ReadRetryPolicy client.Reader requireLeader snapshotStream (Some eventType) ct
         return Snapshot.decode tryDecode events }
 
-    member _.Reload(log, streamName, requireLeader, token, tryDecode, ct): Task<StreamToken * 'event[]> = task {
+    member _.Reload(log, streamName, requireLeader, token, tryDecode, fold, initial, ct): Task<struct(StreamToken * 'state)> = task {
```
bartelink, line 336: ooh going all out on perf eh

src/Equinox.MessageDb/MessageDb.fs
bartelink: For now best to do a Seq.choose and leave the array change for later

src/Equinox.MessageDb/Equinox.MessageDb.fsproj
```diff
@@ -21,6 +21,7 @@
     </ItemGroup>
 
     <ItemGroup>
+        <PackageReference Include="FSharp.Control.TaskSeq" Version="0.4.0-alpha.1" />
```
bartelink, line 24: oops

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -377,25 +376,25 @@ type AccessStrategy<'event, 'state> =
...
     interface ICategory<'event, 'state, 'context> with
         member _.Load(log, categoryName, streamId, streamName, _maxAge, requireLeader, ct) =
-            fetch initial (loadAlgorithm log categoryName streamId streamName requireLeader ct)
+            loadAlgorithm log categoryName streamId streamName requireLeader ct
         member x.TrySync(log, categoryName, streamId, streamName, ctx, _maybeInit, token, state, events, ct) = task {
             let encode e = codec.Encode(ctx, e)
             let encodedEvents: IEventData<EventBody>[] = events |> Array.map encode
             match! context.TrySync(log, categoryName, streamId, streamName, token, encodedEvents, ct) with
             | GatewaySyncResult.Written token' ->
-                let state' = fold state (Seq.ofArray events)
```
bartelink, line 398: this changes the contact, better to leave for the PR that does that

nordfjord: fixed

bartelink:
Thanks, this is going to be a very boring read at the end!

PR #410, opened 2023-07-25
feat!(Stores): Category ctors take name + caching argument goes last

bartelink:
Following @nordfjord and @everzet's leadership in https://github.com/nordfjord/equinox-js/pull/6

- all `*StoreCategory` types now have a `Name` as a second argument rather than having it littering the `Decider.resolve` partial application chain.
- the `caching` argument moves to the end of the constructor argument list, to reflect that Caching, which was once a store-specific integration, is now a fairly separable decorator (one can imagine adjusting it so that it gets applied independent of the `Category` construction at some point)

IME This has made factory wiring in apps much more terse, with less juggling of compiler errors, and also removes the ugly temporary `Equinox.Factory` name, so I trust that on balance this is a good change

### DIFF
--- a/src/Equinox.Core/Category.fs
+++ b/src/Equinox.Core/Category.fs
@@ -32,42 +32,39 @@ open System.Threading.Tasks
 /// Store-agnostic baseline functionality for a Category of 'event representations that fold to a given 'state
 [<NoComparison; NoEquality>]
 type Category<'event, 'state, 'context>
-    (   resolveInner: string -> string-> struct (Core.ICategory<'event, 'state, 'context> * string * (CancellationToken -> Task<unit>) voption),
+    (   name,
+        resolveStream: string -> string -> struct (Core.ICategory<'event, 'state, 'context> * string * (CancellationToken -> Task<unit>) voption),
         empty: struct (Core.StreamToken * 'state)) =
+    member val private Name = name
     /// Provides access to the low level store operations used for Loading and/or Syncing updates via the Decider
-    /// (Normal usage is via the adjacent `module Decider` / `Factory.Resolve` helpers)
-    member _.Stream(log: Serilog.ILogger, context: 'context, categoryName, streamId) =
-        let struct (inner, streamName, init) = resolveInner categoryName streamId
+    /// (Normal usage is via the adjacent `module Decider` / `Stream.Resolve` helpers)
+    member _.Stream(log: Serilog.ILogger, context: 'context, streamId) =
+        let struct (inner, streamName, init) = resolveStream name streamId
         { new Core.IStream<'event, 'state> with
             member _.Name = streamName
             member _.LoadEmpty() =
                 empty
             member _.Load(maxAge, requireLeader, ct) = task {
                 use act = source.StartActivity("Load", ActivityKind.Client)
-                if act <> null then act.AddStream(categoryName, streamId, streamName).AddLeader(requireLeader).AddStale(maxAge) |> ignore
-                return! inner.Load(log, categoryName, streamId, streamName, maxAge, requireLeader, ct) }
+                if act <> null then act.AddStream(name, streamId, streamName).AddLeader(requireLeader).AddStale(maxAge) |> ignore
+                return! inner.Load(log, name, streamId, streamName, maxAge, requireLeader, ct) }
             member _.Sync(attempt, token, originState, events, ct) = task {
                 use act = source.StartActivity("Sync", ActivityKind.Client)
-                if act <> null then act.AddStream(categoryName, streamId, streamName).AddSyncAttempt(attempt) |> ignore
+                if act <> null then act.AddStream(name, streamId, streamName).AddSyncAttempt(attempt) |> ignore
                 let log = if attempt = 1 then log else log.ForContext("attempts", attempt)
-                return! inner.Sync(log, categoryName, streamId, streamName, context, init, token, originState, events, ct) } }
-
-[<AbstractClass; Sealed>]
-type private Stream private () =
-    static member Resolve(cat: Category<'event, 'state, 'context>, log, context): System.Func<string, Core.StreamId, Core.IStream<'event, 'state>> =
-        System.Func<string, Core.StreamId, _>(fun categoryName streamId -> cat.Stream(log, context, categoryName, Core.StreamId.toString streamId))
+                return! inner.Sync(log, name, streamId, streamName, context, init, token, originState, events, ct) } }
 
 [<AbstractClass; Sealed; System.Runtime.CompilerServices.Extension>]
-type Factory private () =
+type Stream private () =
     [<System.Runtime.CompilerServices.Extension>]
-    static member Resolve(cat: Category<'event, 'state, 'context>, log, context): System.Func<string, Core.StreamId, DeciderCore<'event, 'state>> =
-         System.Func<_, _, _>(fun c s -> Stream.Resolve(cat, log, context).Invoke(c, s) |> DeciderCore<'event, 'state>)
+    static member Resolve(cat: Category<'event, 'state, 'context>, log, context): System.Func<Core.StreamId, DeciderCore<'event, 'state>> =
+         System.Func<_, _>(fun sid -> cat.Stream(log, context, Core.StreamId.toString sid) |> DeciderCore<'event, 'state>)
     [<System.Runtime.CompilerServices.Extension>]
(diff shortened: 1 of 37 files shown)


src/Equinox.Core/Category.fs
```diff
@@ -32,41 +32,39 @@ open System.Threading.Tasks
 /// Store-agnostic baseline functionality for a Category of 'event representations that fold to a given 'state
 [<NoComparison; NoEquality>]
 type Category<'event, 'state, 'context>
-    (   resolveInner: string -> string-> struct (Core.ICategory<'event, 'state, 'context> * string * (CancellationToken -> Task<unit>) voption),
+    (   name,
+        resolveStream: string -> string -> struct (Core.ICategory<'event, 'state, 'context> * string * (CancellationToken -> Task<unit>) voption),
         empty: struct (Core.StreamToken * 'state)) =
+    member val private Name = name
```
bartelink, line 38: should this be public or deleted ?

src/Equinox.Core/Category.fs
```diff
@@ -32,41 +32,39 @@ open System.Threading.Tasks
...
+        Stream.Resolve(cat, log, ())
 
 module Decider =
-    let resolveWithContext log (cat: Category<'event, 'state, 'context>) categoryName context streamId =
-        Factory.Resolve(cat, log, context).Invoke(categoryName, streamId) |> Decider<'event, 'state>
-    let resolve log (cat: Category<'event, 'state, unit>) categoryName streamId =
-        resolveWithContext log cat categoryName () streamId
+    let forContext log (cat: Category<'event, 'state, 'context>) context streamId =
+        Stream.Resolve(cat, log, context).Invoke(streamId) |> Decider<'event, 'state>
+    let forStream log (cat: Category<'event, 'state, unit>) streamId =
```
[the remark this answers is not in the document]
nordfjord: I think `for` is better 🤔

src/Equinox.Core/Category.fs
nordfjord: what would you use it for?

src/Equinox.Core/Category.fs
```diff
@@ -32,41 +32,39 @@ open System.Threading.Tasks
...
             member _.Load(maxAge, requireLeader, ct) = task {
                 use act = source.StartActivity("Load", ActivityKind.Client)
-                if act <> null then act.AddStream(categoryName, streamId, streamName).AddLeader(requireLeader).AddStale(maxAge) |> ignore
-                return! inner.Load(log, categoryName, streamId, streamName, maxAge, requireLeader, ct) }
-            member _.TrySync(attempt, (token, originState), events, ct) = task {
-                use act = source.StartActivity("TrySync", ActivityKind.Client)
-                if act <> null then act.AddStream(categoryName, streamId, streamName).AddSyncAttempt(attempt) |> ignore
+                if act <> null then act.AddStream(name, streamId, streamName).AddLeader(requireLeader).AddStale(maxAge) |> ignore
+                return! inner.Load(log, name, streamId, streamName, maxAge, requireLeader, ct) }
+            member _.Sync(attempt, token, originState, events, ct) = task {
```
nordfjord, line 51: Ooh, I love the rebrand <3

nordfjord (approved):
This is awesome!

src/Equinox.Core/Category.fs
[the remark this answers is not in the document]
bartelink: Enough better to use `For` ? :P

src/Equinox.Core/Category.fs
bartelink: Debugging or logging ? I would personally never use it

src/Equinox.Core/Category.fs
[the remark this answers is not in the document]
nordfjord: No 🙅 

what's wrong with `resolveStream`?

src/Equinox.Core/Category.fs
nordfjord: don't expose it then 😅 one less thing to BC break

bartelink: BC?

nordfjord: Backwards Compatibility

bartelink: yes, good point. Just setting people up for disappointment and grumbling!

src/Equinox.Core/Category.fs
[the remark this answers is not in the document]
bartelink: sorry, scanned that too quick - are you suggesting `resolveStream` instead of `forStream`?
While there is a resolution of cat+streamId to container/table + streamName, this just uses the `Stream` on the Cat, making a `Decider` based on that stream?

src/Equinox.Core/Category.fs
```diff
@@ -32,42 +32,38 @@ open System.Threading.Tasks
...
+    (   name,
+        resolveStream: string -> string -> struct (Core.ICategory<'event, 'state, 'context> * string * (CancellationToken -> Task<unit>) voption),
         empty: struct (Core.StreamToken * 'state)) =
     /// Provides access to the low level store operations used for Loading and/or Syncing updates via the Decider
-    /// (Normal usage is via the adjacent `module Decider` / `Factory.Resolve` helpers)
-    member _.Stream(log: Serilog.ILogger, context: 'context, categoryName, streamId) =
-        let struct (inner, streamName, init) = resolveInner categoryName streamId
+    /// (Normal usage is via the adjacent `module Decider` / `Stream.Resolve` helpers)
+    member _.Stream(log: Serilog.ILogger, context: 'context, streamId) =
+        let struct (inner, streamName, init) = resolveStream name streamId
```
nordfjord, line 41: Now's your chance to drop `init` too 😈

src/Equinox.Core/Category.fs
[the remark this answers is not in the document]
nordfjord: 🤔 , yeah, I think `forStream` is the way to go.

`Decider.forStream yuckyLog cat streamId` reads well

src/Equinox.Core/Category.fs
nordfjord: reasoning: it's cosmos specific and amounts to the cosmos category passing itself a value through the core category

bartelink: I have a failure of imagination as to how to achieve that though..

The objective is for the first append per container in any process to do the `CreateStoreProcIfNotExists` (unless you opt out, which would mean you'd need to guarantee to run an `eqx init` has happened for every container every time the stored proc code changes/is versioned)

Also, if there are concurrent first appends to multiple streams (across multiple containers), the inits happen exactly once per container 

(aside from latency, a null stored proc updated costs 6 RU I believe)

nordfjord: > (unless you opt out, which would mean you'd need to guarantee to run an eqx init has happened for every container every time the stored proc code changes/is versioned)

I assumed this was the recommended way?

nordfjord: that said, it sounds like you want an async lazy somewhere?

src/Equinox.Core/Category.fs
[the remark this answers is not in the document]
nordfjord: > Which, depending on the arg order of ctx vs sid is either (right now in this PR as it stands)

```ts
Equinox.Decider.forContext Metrics.log cat
```

Weirdly enough, this is what it looks like in my code, because in my codebase we always pass context first

src/Equinox.Core/Category.fs
nordfjord:
```fsharp
module Async =
    let heinously a = Async.RunSynchronously(a)
    let hronously ct a = Async.Start(a, ct)
```

I'm so here for this 🍿

src/Equinox.Core/Category.fs
nordfjord: https://github.com/jet/equinox/commit/bd8a771eff4eb0e71bfec9b09544401a2f438faf spiked it, it works, let's do it 💪

src/Equinox.Core/Category.fs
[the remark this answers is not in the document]
bartelink: >Weirdly enough, this is what it looks like in my code, because in my codebase we always pass context first
So what's the full sugar, where do you add the `streamId` mapping?

src/Equinox.Core/Category.fs
bartelink: 🤣

PR #415, opened 2023-08-02
refactor(StreamId): Use arrays to pass stream id fragments

nordfjord:


bartelink:
Thanks, I guess this is for the best!

PR #417, opened 2023-08-03
feat(Stores): Make caching and access mandatory

bartelink:
While the value of caching is most debatable for ESDB, in general forcing people to be explicit about their caching strategy is a nudge that's too important to give up

- It also makes the Cosmos and Dynamo implementations cleaner to follow
- Obviates the need for #399 (as Cosmos and Dynamo no longer need specific mapping of available strategies)

Following the logic to its conclusion, considering the Access Strategy should also not be considered a luxury for a non-top system.
- Adds an `Unoptimized` case for EventStore, EventStoreDb, MessageDb and SqlStreamStore 

HT @nordfjord for driving the suggestion

### DIFF
--- a/src/Equinox.Core/Cache.fs
+++ b/src/Equinox.Core/Cache.fs
@@ -101,7 +101,7 @@ type Cache private (inner: System.Runtime.Caching.MemoryCache) =
     member val Inner = inner
 
 type [<NoComparison; NoEquality; RequireQualifiedAccess>] CachingStrategy =
-    /// Do not apply any caching strategy for this Stream.
+    /// Do not apply any caching strategy for this Category.
     | NoCaching
     /// Retain a single 'state per streamName.
     /// Each cache hit for a stream renews the retention period for the defined <c>window</c>.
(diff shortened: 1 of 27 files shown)


nordfjord (approved):
Love it! Next we'll do mandatory access strategies 😉

src/Equinox.MessageDb/MessageDb.fs
```diff
@@ -415,9 +415,9 @@ type MessageDbCategory<'event, 'state, 'context> internal (resolveInner, empty)
     inherit Equinox.Category<'event, 'state, 'context>(resolveInner, empty)
     new(context: MessageDbContext, codec: IEventCodec<_, _, 'context>, fold, initial,
         // For MessageDb, caching is less critical than it is for e.g. CosmosDB
```
nordfjord, line 417:
```
        // While not necessary to control costs, caching can improve the throughput of your application
        // a few times over, as such you should only skip it if you know what you're doing
```

bartelink:
> Love it! Next we'll do mandatory access strategies

rebase didn't even leave that as optional...

src/Equinox.MessageDb/MessageDb.fs
bartelink: Threw it in, there's too much now though...

nordfjord:
Merge it!

src/Equinox.Core/Cache.fs
```diff
@@ -101,6 +101,8 @@ type Cache private (inner: System.Runtime.Caching.MemoryCache) =
     member val Inner = inner
 
 type [<NoComparison; NoEquality; RequireQualifiedAccess>] CachingStrategy =
+    /// Do not apply any caching strategy for this Stream.
+    | NoCaching
```
bartelink, line 105: aka Kimi Raikkonen mode

nordfjord: "The Iceman needs no cache"

bartelink: Indeed. (FWIW ref in my ind was "Leave me alone, I know what I'm doing").
`CachingStrategy.Kimi`

PR #419, opened 2023-08-05
feat!: Push StreamId out to FsCodec

bartelink:
See also https://github.com/jet/FsCodec/pull/100
- first cut done
- [x] removed Cosmos/Dynamo wacky mapping overrides (strong types make wacky hack code stick out a mile!)
- [x] move Tracing and Category into Equinox (drags in FsCodec and Serilog)
- [x] move `Impl` from top of Decider.fs to `Stream.fs` (renamed from Core.fs)
- [x] review if anything should be added to FsCodec (don't want to put any extensions in Equinox/Equinox.Core)
- [x] considered dragging in Cache and Caching, but that drags in AsyncCacheCell too (and System.MemoryCache ref). Thoughts ? Ultimately there are no interesting stores that don't ref it?
- [x] Apply `module Stream` pattern as per FsCodec throughout samples/integration tests
cc @nordfjord


src/Equinox/Category.fs
```diff
@@ -20,24 +20,22 @@ type ICategory<'event, 'state, 'context> =
...
 [<NoComparison; NoEquality>]
-type Category<'event, 'state, 'context>(categoryName, resolveStream) =
-
-    /// Stores without custom routing for categoryName/streamId to Table/Container etc use this default impl
-    new(categoryName, inner) = Category(categoryName, fun streamId -> struct (inner, Core.StreamName.render categoryName streamId))
-
-    /// Provides access to the low level store operations used for Loading and/or Syncing updates via the Decider
-    /// (Normal usage is via the adjacent `module Decider` / `Stream.Resolve` helpers)
+type Category<'event, 'state, 'context>(categoryName, inner: Core.ICategory<'event, 'state, 'context>) =
     member _.Stream(log: Serilog.ILogger, context: 'context, streamId: string) =
```
nordfjord, line 36:
```suggestion
    member _.Stream(log: Serilog.ILogger, context: 'context, streamId: FsCodec.StreamId) =
```

Not sure if possible or not

bartelink: Good call, thanks!
StreamName manipulation for MDB snapshot (grep `let inline streamName`) is a big ugly, but no special cases or direct UMX usage anywhere now

PR #422, opened 2023-08-12
fix!: Remove Async from TransactAsync, TransactExAsync

bartelink:
Another run at #314
Having >10 overloads of transact has pros and cons; the seriousness of the cons is largely dependent on the IDE

While the single change I had to make to the consumption in this repo seems innocuous, IME it can be excruciating to resolve when you're just trying to make something work.

So I'm still not sure if it should actually be done...

### DIFF
--- a/samples/Store/Domain/Favorites.fs
+++ b/samples/Store/Domain/Favorites.fs
@@ -82,7 +82,8 @@ type Service internal (resolve: ClientId -> Equinox.Decider<Events.Event, Fold.S
     // NOTE not a real world example - used for an integration test; TODO get a better example where it's actually relevant
     member _.UnfavoriteWithPostVersion(clientId, sku) =
         let decider = resolve clientId
-        decider.TransactEx((fun c -> (), decideUnfavorite sku c.State), fun () c -> c.Version)
+        let mapResult () (c: Equinox.ISyncContext<_>) = c.Version
+        decider.TransactEx((fun c -> (), decideUnfavorite sku c.State), mapResult)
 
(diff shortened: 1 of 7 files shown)


samples/Infrastructure/Store.fs
```diff
@@ -97,9 +97,9 @@ module Cosmos =
         member x.TipMaxJsonLength =     p.GetResult(TipMaxJsonLength, 30_000)
         member x.QueryMaxItems =        p.GetResult(QueryMaxItems, 10)
 
-    let logContainer (log: ILogger) name (mode, endpoint, db, container) =
```
bartelink, line 100: this stuff is just sitting here - will likely make a PR with some tweaks inc this on the road to completing https://github.com/jet/dotnet-templates/pull/131

samples/Store/Domain/Favorites.fs
```diff
@@ -82,7 +82,8 @@ type Service internal (resolve: ClientId -> Equinox.Decider<Events.Event, Fold.S
     // NOTE not a real world example - used for an integration test; TODO get a better example where it's actually relevant
     member _.UnfavoriteWithPostVersion(clientId, sku) =
         let decider = resolve clientId
-        decider.TransactEx((fun c -> (), decideUnfavorite sku c.State), fun () c -> c.Version)
+        let mapResult () (c: Equinox.ISyncContext<_>) = c.Version
```
bartelink, line 85: I doubt this is an obvious solution to anyone new to F# - the error message does mention 'a type annotation'

PR #426, opened 2023-08-18
chore(MsSql): Update MSSQL image workaround

bartelink:
Applies the workaround suggested in https://github.com/microsoft/mssql-docker/issues/2#issuecomment-1683956282


bartelink:
Seems to work, thanks for sharing!
(but my test env is down so won't be merging this yet...)

PR #427, opened 2023-08-21
feat(Batcher): Add Limiter option

bartelink:
Adds a facility to the Batcher that (optionally) extends the Linger period to include a phase during which a `SemaphoreSlim` is acquired:
- the semaphore is `Release`d when the processing has finished
- further items can join the lingering batch during the acquisition period


nordfjord (approved):
LGTM, it still surprises me that his stuff isn't available in the stdlib 🥴

bartelink:
It definitely feels like a cleaner impl should be possible, and I'm also surprised I have not ran into anything similar.
Being able to leave it open while there's batches in flight is definitely a big upgrade on trying 5/10/50/100 and convincing oneself that such a linger time makes sense overall in the context of some anticipated load

PR #431, opened 2023-09-14
feat(Cosmos): TrustLocalEmulator=true

bartelink:
Provides a lean and safe one-stop-shop way to use a local Cosmos emulator by setting `EQUINOX_COSMOS_CONNECTION` to `TrustLocalEmulator=true` instead of one or more of:
1. having risky and/or messy conditional logic in your system that can ignore SSL cert checking
2. having to copy around magic values for the Emulator's URI or AccessKey
3. having to (on a Mac) register the Emulator's self-signed certificate in the Key Chain in order to trust it

----

Sadly this doesn't yet work on Mac so is useless as it stands, as `docker-compose-cosmos.sh` hangs when creating the database:

1. could be due to ports used in `Direct` mode in Docker not being exposed correctly
3. could be due to some other aspect of the system needing the Emulator Cert to be registered in the KeyChain

(the MS examples for skipping TLS all show `Gateway` mode)

### DIFF
--- a/README.md
+++ b/README.md
@@ -681,6 +681,10 @@ NOTE There's [no Apple Silicon emulator available as yet](https://github.com/Azu
 NOTE Have not tested with the Windows Emulator, but it should work with analogous steps.
 
 ```bash
+# magic connection string value that CosmosStoreConnector supports to
+# a) avoid having to copy values around
+# b) having to add code to trust cert and/or assent to elevation primpts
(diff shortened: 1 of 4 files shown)


bartelink:
Superseded by https://github.com/Azure/azure-cosmos-dotnet-v3/issues/4222
https://github.com/jet/equinox/pull/443

Will circle back and see if it actually becomes possible to stop trusting the cert entirely
The fact this PR did not work suggests there's still something missing from the equation when using Direct mode against a dockerized emulator

PR #434, opened 2023-12-30
feat(Cosmos): eqx query

bartelink:
If we make `eqx init` also index the following paths:
- `"/u/[]/c/?"`
- `"/u/[]/d/*"`

We can:
- pay a little more on writes to maintain 2 more indexed things
- add uncompressed unfolds (so we can query against `u.d.<FieldName>`)
- identify streams to read or transact against by category name + an unfold name + arbitrary criteria targeting uncompressed unfolds for 2.8RU
- pull back `u` (and, optionally `p`) and provide a paged result based on hydrating the State from the (potentially compressed) Snapshot and rendering that (can convey Version and State)
  - `p` lets you identify the stream, but that's optional if it's state is sufficient
- pull back `u`, `p`, `_etag` and then (potentially concurrently) do a `Transact` read and/or write roundtrip based on only that input
  - could also do a null update to freshen the Cache
  
Yes, there's other cleanup junk in this PR


src/Equinox.CosmosStore/CosmosStore.fs
```diff
@@ -625,7 +625,8 @@ module internal Query =
         let mutable i = 0
         while query.HasMoreResults do
             let! t, (res: FeedResponse<'t>) = (fun ct -> query.ReadNextAsync(ct)) |> Stopwatch.time ct
-            yield map i t res }
+            yield map i t res
+            i <- i + 1 }
```
bartelink, line 629: this fixes values in the logs, and makes the maxbatches limit be enforced correctly

nordfjord:
Would be good to see an example of how to use it in actual applications as well

bartelink:
Yup - the overview is also in need of a complete rewrite first though...

bartelink:
## Learnings/decisions compared to original

- Main difference/learning/variance is that providing a function to generate a memento means no TransactMany/QueryMany API variant.
- Searching based on `u/[]/c` means no need to filter based on `id = "-1"`
- `eqx query` shows 3 modes (inc retrieving the `p` values); How to use the `p` value is an external concern

## Original proposal:

We can:
- pay more on writes to maintain 2 more indexed things (affects write ops, but there's only one per stream and the first path should remain stable)
  - no, have not assessed the damage in terms of write costs
  - note that the `/d/*` means only data in uncompressed folds, so should be relatively low
- add uncompressed unfolds (so we can query against `u.d.<FieldName>`)
- identify streams to read or transact against by category name + an unfold name + arbitrary criteria targeting uncompressed unfolds for 2.8RU
- `QueryMany`: pull back `u` (and, optionally `p`) and provide a paged result based on hydrating the State from the (potentially compressed) Snapshot and rendering that (can convey Version and State)
  - `p` lets you identify the stream, but that's optional if it's state is sufficient
- `TransactMany`: pull back `u`, `p`, `_etag` and then (potentially concurrently) do a `Transact` read and/or write roundtrip based on only that input
  - could also do a null update to freshen the Cache
- for streams that are `AccessStrategy.Unoptimized` (events only), it would also work
   - would have to get `e` in lieu of `u`
   - as there is no `u`, there's no ability to filter based on uncompressed folds
   - would also need to filter by `id = "-1"` (to rule out other event batches), or retrieve those too and sort by `c.i` (if you know you'll hydrate the bulk of them)

PR #436, opened 2023-12-31
feat(Cosmos): compressUnfolds -> shouldCompress

bartelink:
Generalizes the compressUnfolds option (which defaults to `true`), to allow control at the specific Unfold event level.

Omitting continues to default to compressing (and not indexing) Unfolds.


nordfjord (approved):
lgtm

PR #438, opened 2024-01-08
refactor(Cosmos): SyncExp

bartelink:
Preparatory refactoring in advance of #434
- `FromI` was always badly named
- `SyncExp` came later; there's definitely room to do more but the main goal here is to do 434 without inducing risk

### DIFF
--- a/src/Equinox.CosmosStore/CosmosStore.fs
+++ b/src/Equinox.CosmosStore/CosmosStore.fs
@@ -124,14 +124,14 @@ type Tip = // TODO for STJ v5: All fields required unless explicitly optional
 [<NoComparison; NoEquality>]
 type Position = { index: int64; etag: string option }
 module internal Position =
-    /// NB very inefficient compared to FromEtagAndNext or using one already returned to you
-    let fromI (i: int64) = { index = i; etag = None }
-    /// If we have strong reason to suspect a stream is empty, we won't have an etag (and Writer Stored Procedure special cases this)
-    let fromKnownEmpty = fromI 0L
-    /// Blind append mode
-    let fromAppendAtEnd = fromI -1L // sic - needs to yield value -1 to trigger stored proc logic
     let fromEtagOnly (value: string) = { index = -2; etag = Some value }
-    let fromEtagAndNext (etag, n) = { index = n; etag = Some etag }
+    let fromEtagAndIndex (etag, n) = { index = n; etag = Some etag }
+    /// NB a token without an etag is inefficient compared to fromEtagAndIndex, so paths to this should be minimized
+    let fromIndex (x: int64) = { index = x; etag = None }
+    /// If we have strong reason to suspect a stream is empty, we won't have an etag (and Writer Stored Procedure special cases this)
+    let fromKnownEmpty = fromIndex 0L
+    /// Blind append mode
+    let fromAppendAtEnd = fromIndex -1L // sic - needs to yield value -1 to trigger stored proc logic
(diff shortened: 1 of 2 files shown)


src/Equinox.CosmosStore/CosmosStore.fs
```diff
@@ -615,7 +622,18 @@ module internal Tip =
         | ReadResult.NotFound -> return Result.NotFound
         | ReadResult.Found tip ->
             let minIndex = maybePos |> Option.map (fun x -> x.index)
-            return Result.Found (Position.fromTip tip, tip.i, Enum.EventsAndUnfolds(tip, ?maxIndex = maxIndex, ?minIndex = minIndex) |> Array.ofSeq) }
+            return Result.Found (Position.fromEtagAndIndex (tip._etag, tip.n), tip.i, Enum.EventsAndUnfolds(tip, ?maxIndex = maxIndex, ?minIndex = minIndex) |> Array.ofSeq) }
+    let tryFindOrigin (tryDecode: ITimelineEvent<EventBody> -> 'event voption, isOrigin: 'event -> bool) xs =
```
bartelink, line 626: pulled out as we'll use it in #434

src/Equinox.CosmosStore/CosmosStore.fs
```diff
@@ -694,16 +712,11 @@ module internal Query =
...
-            | ValueNone -> false
-            | ValueSome e ->
-                items.Insert(0, e) // WalkResult always renders events ordered correctly - here we're aiming to align with Enum.EventsAndUnfolds
-                isOrigin e
-        let f, e = xs |> Seq.tryFindBack isOrigin' |> Option.isSome, items.ToArray()
-        { found = f; maybeTipPos = Some pos; minIndex = i; next = pos.index + 1L; events = e }
+    let scanTip (tryDecode, isOrigin) (pos: Position, i: int64, xs: #ITimelineEvent<EventBody>[]): ScanResult<'event> =
+        let ok, e = xs |> Tip.tryFindOrigin (tryDecode, isOrigin) |> function
+            | ValueSome (_, events) -> true, events
+            | ValueNone -> false, Array.empty
```
nordfjord, line 718: maybe `tryFindOrigin` should return a `bool * 'E[]`?

src/Equinox.CosmosStore/CosmosStore.fs
```diff
@@ -615,7 +622,18 @@ module internal Tip =
...
+        let stack = ResizeArray()
+        let isOrigin' (u: ITimelineEvent<EventBody>) =
+            match tryDecode u with
+            | ValueNone -> false
+            | ValueSome e ->
+                stack.Insert(0, e) // WalkResult always renders events ordered correctly - here we're aiming to align with Enum.EventsAndUnfolds
+                isOrigin e
+        match xs |> Seq.tryFindBack isOrigin' with
+        | Some x -> ValueSome (x, stack.ToArray())
+        | None -> ValueNone
```
nordfjord, line 636:
```suggestion
        | Some x -> (true, stack.ToArray())
        | None -> (false, [||])
```

nordfjord (approved):

src/Equinox.CosmosStore/CosmosStore.fs
bartelink: Hm #434 makes it less obvious as we use the value in another case
also if it wasnt for the ValueSome vs Some, there'd be less debate

src/Equinox.CosmosStore/CosmosStore.fs
bartelink: As per other comment, we use it in the next PR

PR #441, opened 2024-01-15
docs: Remove Command/Handler

bartelink:
Some of the samples are showing their age...
- There was once a `Handler` that went with the `Service` - while this did separate out technical concerns from Service concerns, the cost of another concept ultimately is not justifiable
- `execute` and `query` helpers are similarly well intentioned and arguably separate things out technical concerns from business logic, but again don't bring enough to justify naming them
- once you lose those technical abstractions, the Command Pattern ceases to add anything

### DIFF
--- a/samples/TodoBackend/Todo.fs
+++ b/samples/TodoBackend/Todo.fs
@@ -36,7 +36,7 @@ module Fold =
     let private evolve s = function
         | Events.Added item ->    { items = item :: s.items; nextId = s.nextId + 1 }
         | Events.Updated value -> { s with items = s.items |> List.map (function { id = id } when id = value.id -> value | item -> item) }
-        | Events.Deleted          { id=id } -> { s with items = s.items  |> List.filter (fun x -> x.id <> id) }
+        | Events.Deleted          { id = id } -> { s with items = s.items  |> List.filter (fun x -> x.id <> id) }
         | Events.Cleared ->       { s with items = [] }
         | Events.Snapshotted      { items = items } -> { s with items = List.ofArray items }
     let fold = Array.fold evolve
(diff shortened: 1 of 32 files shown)


samples/TodoBackend/Todo.fs
```diff
@@ -34,16 +34,16 @@ module Fold =
...
         | Events.Snapshotted      { items = items } -> { s with items = List.ofArray items }
     let fold = Array.fold evolve
 
 type Command = Add of Events.Todo | Update of Events.Todo | Delete of id: int | Clear
 
-let interpret c (state: Fold.State) = [|
+let decide c (state: Fold.State) = [|
     match c with
     | Add value -> Events.Added { value with id = state.nextId }
     | Update value ->
```
nordfjord, line 49: You know... if you weren't using the command pattern you could return the patched item as a result from the decision 😜

nordfjord (approved):
Much better!

samples/TodoBackend/Todo.fs
bartelink: the tryfind roundtrip is mandated by the Todobackend spec/protocol
But i did get rid of the Command and it's the better for it

bartelink: Most commands are now removed. I added a TODO to the docs where there's a section laying out the Command handling for Favorites. I need to decide whether to do a compare and contrast, or simply show the recommended solution (skip the Command stuff)

nordfjord: > skip the Command stuff

You already have a link to Jeremie's article, so it should be fine to avoid the commands entirely.

bartelink: Good point - the world is definitely sorted for examples of shoehorning everything into Commands
And thanks to the ChatGPT factor, nobody's going to ever forget how well it and Mediatr solve the problems of the world

PR #452, opened 2024-04-10
fix(Cache): Correct wait for concurrent request

bartelink:
Fixes the condition for the swap, which has a race condition that can mean delivering the value we've just literally determined to be too old
Reduce a test timeout to hopefully reduce flakiness
Also fixes an oversight: even if a concurrent request was too old for the `maxAge`, it can still serve as the base state for an incremental load (not a big deal in the normal case; only the very first load would be suboptimal)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ The `Unreleased` section name is replaced by the expected version of next releas
 ### Removed
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


bartelink:
And yes, the tests should not be using Delays, and instead using `TimeProvider` in most cases, in order to make them quicker, more deterministic and correct (would really apprececiate a PR regarding that; it's unfortunately too far down my overall list of priorities at the moment...)

PR #453, opened 2024-05-06
feat(CosmosStore): Group metrics by Category, add min/max RU, tidy

bartelink:
Follow-up to #449
- Separate `Tip` operation from other `Read` activities
- Add min/max RU costs per operation
- Allow opt-in grouping of CosmosStore InternalMetrics by category
- Formatting improvements


bartelink:
The rate metrics would probably be useful on the DynamoStore side too
Arguably Tables can be managed more granularly there, so having the option to break down by category might be of less value.

PR #454, opened 2024-05-07
feat(Core.Batcher): Settable Linger

bartelink:


bartelink:
Logically this is 4.0.4 and triggers a release of Equinox.Core (only)

PR #467, opened 2024-09-07
feat(Cosmos): Use Azure.Cosmos STJ impl

bartelink:
- Use the in-the-box `System.Text.Json` support provided in `Microsoft.Azure.Cosmos`
- `CosmosStoreConnector`: Remove `requestTimeout` mandatory argument - the particulars of tuning that are a lower level concern best left to the SDK team to manage (they've adjusted it from 5 to 6 seconds internally, and their retry logic is tuned and managed in accordance with the CosmosDB Service's SLAs and/or behavior under stress to a degree that neither Equinox or applications have much business concerning themselves with)
  - NOTE the parameter is still exposed for backcompat, and it continues to be something you want to consider taking ownership of if your calls/queries affect the user experience


src/Equinox.CosmosStore/Equinox.CosmosStore.fsproj
```diff
@@ -27,7 +27,7 @@
     <PackageReference Include="FSharp.Core" Version="6.0.7" ExcludeAssets="contentfiles" />
 
     <PackageReference Include="FSharp.Control.TaskSeq" Version="0.4.0" />
-    <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.35.4" />
+    <PackageReference Include="Microsoft.Azure.Cosmos" Version="3.43.1" />
```
bartelink, line 30: update all to max 3.43 ver

PR #471, opened 2024-12-20
Recommend a data store in README

Happypig375:


bartelink:
Will come back to this in a few days; the problem for me is that there's a lot of context that feeds into a database choice, so having a statement like this doesn't sit well. But I do agree that having a quickstart "if you're just trying to see how it fits for you and you just want a decision, any decision, pick this".... being missing as you suggest

PR #473, opened 2025-02-02
feat(eqx): Add top, destroy, -Q, -sl etc

bartelink:
- Adds `eqx top`, which summarizes categories and/or stream stats (see README for examples)
- Adds `eqx destroy`, supporting deleting streams (and/or categories thereof) from an Equinox store
- `eqx dump`: Applies a bare mode when in `-Q`uiet mode
- Various improvements to eqx tool outputs and cleanups of flags

### DIFF
--- a/tools/Equinox.Tool/CosmosStoreLinq.fs
+++ b/tools/Equinox.Tool/CosmosStoreLinq.fs
@@ -1,0 +1,42 @@
+namespace Equinox.CosmosStore.Linq
+
+open Equinox.Core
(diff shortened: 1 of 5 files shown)


tools/Equinox.Tool/CosmosStoreLinq.fs
```diff
@@ -0,0 +1,42 @@
+namespace Equinox.CosmosStore.Linq
```
bartelink, line 1: will move to Equinox.CosmosStore in PR 450

PR #474, opened 2025-02-03
chore(DynamoStore)!: Handle breaking TransactWriteItems API change

bartelink:
Handle breaking changes to the `TransactWriteItems` wrapping in `FSharp.AWS.DynamoDB` support required to support multiple-Table transactions


bartelink:
Seems a bit uglier here with this approach - the DU was actually useful for handling the special case of single request transacitons with pattern matching @purkhusid @samritchie

Issue #475, opened 2025-06-10
Add read retry policy to EventStoreDb


DmitryBatalov:
I was testing the ESDB cluster and encountered an issue with failover behavior. When I removed a follower node, the connection did not automatically switch to a healthy follower node and instead continued attempting to connect to the inactive one. I believe this functionality should ideally be handled by the ESDB .NET client, however, as a temporary solution, I would like to implement a retry policy on the Equinox side.

PR #476, opened 2025-06-10
Added read retry policy to EventStoreDb.

DmitryBatalov:
Here is the implementation for the issue #475. 

The issue I was unable to resolve is efficiently loading only the subset of events up to the origin point. In the current implementation, the entire batch is loaded first, and then filtered using "take while origin." This approach has a significant drawback: in the worst-case scenario, substantially more data than necessary will be loaded into memory.


DmitryBatalov:
@bartelink could help me with failing build? Didn't get what's wrong :)

bartelink:
Thanks - I'd have to double check, but as you can appreciate "resilience is more important than loading extra data" is definitely very subjective if you have a long stream with massive events and you were counting on the actual defined behavior (not over-reading) being in place

If there is no test coverage regarding not overreading in the ESDB context, I'm surprised - that constitutes a bug in the test suite (IIRC the integration tests check for batch read log emissions across all stores).

DmitryBatalov:
Closing PR.

PR #491, opened 2026-04-10
chore(EventStore): Rework docker config to single secure node on 23.10.8

bartelink:
Replace the 3-node TLS EventStoreDB cluster (21.10.0-buster-slim) with a single secure node (23.10.8-bookworm-slim) using es-gencert-cli for dev certificates. Remove docker-compose.env and cluster network config.

Update test connection strings:
- gRPC: single-node with auth credentials and tlsVerifyCert=false
- Legacy TCP: direct URI with DisableServerCertificateValidation()

23.10 LTS is the final version supporting both gRPC and legacy TCP

Medium term future is replacing the 2x .EventStore drivers with a Kurrent one targeting the new SDK


bartelink:
superseded by #493

